PR #2, opened 2018-07-27
make sure SAMPLE_ID isn't used when not existent

thomasyu888:


processing/clinical.py
```diff
@@ -370,8 +370,9 @@ def validate_helper(self, clinicalDF, clinicalSampleDF, oncotreeLink):
 				total_error += "Sample: There can't be any blank values for PATIENT_ID\n"
 
 			#CHECK: within the sample file that the sample ids match the patient ids
-			if not all([patient in sample for sample, patient in zip(clinicalSampleDF[sampleId], clinicalSampleDF[patientId])]):
-				total_error += "Sample: PATIENT_ID's much be contained in the SAMPLE_ID's (ex. SAGE-1 <-> SAGE-1-2)\n"
+			if haveSampleColumn:
```
zaro0508, line 373: do you need this?  doesn't the else already say that there IS a sample column?

zaro0508 (changes requested):

processing/clinical.py
zaro0508: oops.  my mistake.  the github review made it look like this is right under line 247.

zaro0508 (approved):

kkdang (approved):

PR #7, opened 2018-08-16
CNA 

thomasyu888:
Blank values are for genes that were not assessed.  Do not fill blank values with Zero.

### DIFF
--- a/analyses/clinicalData/oncotree_code_converter.py
+++ b/analyses/clinicalData/oncotree_code_converter.py
@@ -88,7 +88,7 @@ def process_clinical_file(oncotree, clinical_filename, spreadsheet_fields):
 				header.append(ONCOTREE_PRIMARY_NODE)
 			if ONCOTREE_SECONDARY_NODE not in header:
 				header.append(ONCOTREE_SECONDARY_NODE)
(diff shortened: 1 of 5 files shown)


thomasyu888:
Do not merge yet, still a bit more changes that need to happen

analyses/clinicalData/oncotree_code_converter.py
```diff
@@ -153,7 +153,7 @@ def process_clinical_file_json(oncotree, clinical_filename):
 			try:
 				data[header.index(CANCER_TYPE_DETAILED)] = cancer_types[CANCER_TYPE_DETAILED]
 			except IndexError:
-				data.append(cancer_types[CANCER_TYPE_DETAILED])
+				data.append(cancer_types[CANCER_TYPE_DETAILED].replace(u"\u2013","-"))
```
thomasyu888, line 156: This had to be added, because of how python 2.7 deals with unicodes.  Code SGTTL
http://oncotree.mskcc.org/api/tumorTypes/tree?version=oncotree_2017_06_21 has a non ascii character as a -

kkdang (approved):

PR #13, opened 2018-08-28
Pypi update

thomasyu888:


thomasyu888:
Do not pull in yet.  There are revisions that need to happen to create the genie package

genie/toRetract.py
```diff
@@ -2,7 +2,7 @@
 import pandas as pd
 import argparse
 import os
-import process_functions
+from . import process_functions
```
kkdang: Fix this

kkdang (approved):

PR #19, opened 2018-09-04
Clinical onco

thomasyu888:
* Remove support for both oncotree in database to staging
* Removes the need for a separate script to add cancer type, and oncotree primary/secondary nodes


thomasyu888:
Do not pull in yet

kkdang (approved):

PR #38, opened 2018-10-19
Python3 conventions, no need to check mutations for duplication

thomasyu888:
* Out of scope error, python3 specific
* no need to check mutation file for duplications (Only one mutation file allowed to be submitted)


xindiguo (approved):
Looks good!

PR #61, opened 2019-01-24
Batch error while processing vcf/maf, need to change path_to_genie

thomasyu888:
```
ERROR: Could not write to output file /root/Genie/MSK/*_vs_NORMAL.vep.vcf
```
This error makes me believe that for whatever reason, vep is not able to write to inside of the docker container, and this makes sense because VEP is mounted into the container while running.

Although can't explain why the bottom works.
```
STATUS: Running VEP and writing to: /root/Genie/JHU/data_mutations_extended_JHU.vep.vcf
```

This PR is a POSSIBLE TEMPORARY FIX- 

This issue will be revisited in the input_to_database refactor by adding a `workdir` parameter


thomasyu888:
This also resolves any issues with filling up the 10GB limit of docker space

kkdang (approved):

PR #72, opened 2019-03-07
Write GENIE release files to .synapseCache/ add vcf2maf dependency

thomasyu888:
`~/.synapseCache` is mounted during the batch GENIE processes. Make use of this mount to write release files to this.


genie/dashboard_table_updater.py
```diff
@@ -515,9 +515,9 @@ def run_dashboard(syn, database_mappingdf, release, staging=False, public=False)
 
 	'''
 	update_release_numbers(syn, database_mappingdf, release = release)
-	update_data_release_file_table(syn, database_mappingdf)
 
 	if not staging and not public:
```
kkdang: Want it to run if public

kkdang (approved):

PR #102, opened 2019-05-13
DEV: Restrict genie package to use python 3

thomasyu888:


.travis.yml
```diff
@@ -10,18 +10,7 @@ before_install:
 
 matrix:
   include:
-    - os: linux
-      python: 2.7.14
     - os: linux
       python: 3.5.3
     - os: linux
       python: 3.6.2
```
kdaily, line 16: Why not Python 3.7?

setup.py
```diff
@@ -9,11 +9,12 @@
       license='MIT',
       packages=find_packages(),
       zip_safe=False,
+      python_requires='>=3.0',
```
kdaily: This should be at least 3.5 since the travis build starts there.

setup.py
```diff
@@ -9,11 +9,12 @@
...
       entry_points={
         'console_scripts': ['genie = genie.__main__:main']},
       install_requires=[
         'pandas>=0.20.0',
         'synapseclient>=1.9',
         'httplib2>=0.11.3',
         'pycrypto>=2.6.1',
-        'PyYAML>=3.11'])
+        'PyYAML>=5.1'])
```
kdaily, line 20: why such a big jump?

kdaily (changes requested):

setup.py
thomasyu888: I was just using an old pyyaml version.  There are new specs about how to load yaml file which made me have to use `5.1` #78

.travis.yml
thomasyu888: Added.

thomasyu888:
@kdaily Fixed.

PR #103, opened 2019-05-15
Fix mut in cis filter, add option 3

thomasyu888:
#65 Had a breaking filter change
* as.numeric on a `level` column doesn't actually turn the values of that column into its numerical values
* Lint
* Add option 3 - Change mutationInCis filter options to be ON, OFF, FLAG
* Add column in maf: mutationInCis_Flag.  True if in list of flagged variants, otherwise blank.


analyses/mergeFlag/test_flag_variants.R
```diff
@@ -0,0 +1,81 @@
...
+                              "HGVSp_Short")]
+  expected$t_alt_count_num = c(1, 2)
+  # Change everything to factors or else its difficult to validate is absurd
+  # This is ok because we don't do anything with factors after the function is ran
+  expected <- data.frame( lapply( expected , factor ))
+  tbl <- data.frame( lapply( tbl , factor ))
+  expect_equal(tbl, expected[,colnames(tbl)])
+})
+
+genieMutData$Start_Position = "15"
```
thomasyu888: Only Update the index.

genie/database_to_staging.py
```diff
@@ -357,14 +363,29 @@ def mutation_in_cis_filter(
...
+    remove_samples = \
+        filtered_samplesdf['Tumor_Sample_Barcode'].drop_duplicates()
+
+    # Find variants to flag
+    variant_flagging = syn.tableQuery(
+        "SELECT Tumor_Sample_Barcode FROM {} where Flag = 'FLAG' and "
+        "Tumor_Sample_Barcode is not null".format(variant_filtering_synId))
+    flag_variantsdf = variant_flagging.asDataFrame()
+
+    flag_variantsdf['removeVariants'] = \
```
thomasyu888: flagged variants

kkdang (approved):

PR #111, opened 2019-05-21
DEV: don't use synu shorthand

kdaily:


thomasyu888:
Why is this a bad thing?  I use `import pandas as pd` all the time. 

* Edit:  `import pandas as pd` has been used for so long that its become very common.

thomasyu888 (approved):

PR #60, opened 2019-01-23
DEV: Refactor input_to_database.py

thomasyu888:
* abstraction of functions
* add unit tests
* Lint
* Creation of `bin/`

PR #112, opened 2019-05-21
remove redundant import

kdaily:


thomasyu888:
`subprocess` is never used in this script.  Removed in #60

PR #116, opened 2019-05-21
Refactor input_to_db -> _check_valid

thomasyu888:
This function basically calls the `validate.validate` function and catches any filename errors.
* Add unit tests


thomasyu888:
This is unnecessary as of #126

Issue #125, opened 2019-05-23
raw_input not part of python 3


kdaily:
https://github.com/Sage-Bionetworks/Genie/blob/develop/genie/validate.py#L58

Issue #108, opened 2019-05-21
do not suggest saving username and password to bashrc


kdaily:
This circumvents Synapse security policies. We should not suggest or support this kind of password saving.

https://github.com/Sage-Bionetworks/Genie/blob/master/genie/validate.py#L52

PR #126, opened 2019-05-24
Refactor input_to_db -> validate.py

thomasyu888:
* move and refactor synapse_login #108 , #125 
* Remove `_check_valid` and include in `validate.py`
* Add unit tests
* Lint
* validate single file workflow:  Determine filetype, get error/warning messages, get valid status
* fix Argparse


genie/validate.py
```diff
@@ -1,125 +1,117 @@
...
+    '''
+    filetype = None
+    # Loop through file formats
+    for file_format in PROCESS_FILES:
         try:
-            PROCESS_FILES[fileType](syn, center, threads).validateFilename(filePath)
-        except AssertionError as e:
-            raise ValueError("Your filename is incorrect!\n%s\nPlease change your filename before you run the validator again."  % e)
-    total_error, warning = PROCESS_FILES[fileType](syn, center, threads).validate(filePathList=filePath, oncotreeLink=oncotree_url, testing=testing, noSymbolCheck=noSymbolCheck)
+            validator = PROCESS_FILES[file_format](syn, center)
```
kdaily: should go outside of the try

genie/validate.py
```diff
@@ -1,125 +1,117 @@
...
+            filetype = validator.validateFilename(filepathlist)
+        except AssertionError:
```
kdaily, line 31: should not assume that it's an assertion error. or, you need to document that `validateFilename` functions must raise an `AssertionError`. I think it should be a `ValueError`.

genie/validate.py
```diff
@@ -1,125 +1,117 @@
...
+        message - error + warning
+    '''
+    # Complete error message
     message = "----------------ERRORS----------------\n"
     if total_error == "":
         message = "YOUR FILE IS VALIDATED!\n"
         logger.info(message)
         valid = True
     else:
         for errors in total_error.split("\n"):
```
kdaily, line 59: add in explicit statement that the file is invalid here.

genie/validate.py
```diff
@@ -1,125 +1,117 @@
...
-        [syn.store(synapseclient.File(path, parent=uploadToSynapse)) for path in filePath]
-    return(message, valid)
-
-def perform_validate(syn, args):    
-    if args.testing:
-        databaseToSynIdMapping = syn.tableQuery('SELECT * FROM syn11600968')
+    return(valid, message)
+
+
+def validate_single_file_workflow(syn,
```
kdaily: `workflow` is a word used to describe other things in this implementation. would advise a different word, or dropping it (`validate_single_file`)

genie/__main__.py
```diff
@@ -1,37 +1,100 @@
...
+        else:
+            syn = synapseclient.login(
+                email=username,
+                password=password,
+                rememberMe=True,
+                silent=True)
+    return(syn)
+
+
+def perform_validate(syn, args):
```
kdaily: Unsure why this genie-specific functionality got moved to the command line only?

kdaily (changes requested):

genie/__main__.py
thomasyu888: Honestly was just unsure what to name the function. I could name it main in validate.py but this is basically the function that invokes the validate.py functions.

genie/validate.py
thomasyu888: The reason it is in a try is because i need to capture the error. But perhaps this try catch should be handled in validate.py and return the right error messages.

kdaily: just the instantiation of the `validator` should go outside of the `try` block (or in it's own). The `AssertionError` is there to catch exceptions from validating the filename, right?

genie/__main__.py
kdaily: I wasn't worried about the name; but now there are two places with GENIE specific functionality (the command line tool and `validate.py`). If you capture it all in `validate.py` as it was it will be easier to turn that into an interface/class of some kind.

genie/validate.py
thomasyu888: I assume there are errors from line 56.  If theres not any errors, message is switched to "YOUR FILE IS VALIDATED"

genie/validate.py
thomasyu888: Ya.. It should most likely be a `ValueError`.. Should be a separate github issue.  Should most likely document this in `_validateFilename` in `example_filetype_format.py`

genie/__main__.py
thomasyu888: Got it.  Will move it back, but keep `synapse_login` in `__main__`

thomasyu888:
@kdaily Can you take a look at this again?

genie/__main__.py
```diff
@@ -1,37 +1,58 @@
...
+    If not saved, then user is prompted username and password.
+
+    :returns:     Synapseclient object
+    """
+    try:
+        syn = synapseclient.login(silent=True)
+    except Exception:
+        if username is None and password is None:
+            raise ValueError(
+                "Please specify --syn_user, --syn_pass the first time "
```
kdaily: Add link to our docs about login options - use keyring, or config, or a 'rememberMe'

genie/__main__.py
```diff
@@ -1,37 +1,58 @@
...
+                "you run this script.")
+        else:
+            syn = synapseclient.login(
+                email=username,
+                password=password,
+                rememberMe=True,
```
kdaily: remove or parameterize `rememberMe`

genie/example_filetype_format.py
```diff
@@ -51,11 +51,12 @@ def read_file(self, filePathList):
     def _validateFilename(self, filePath):
         '''
         Function that changes per file type for validating its filename
+        Expects an assertion error.
 
         Args:
             filePath: Path to file
         '''
-        pass
+        assert True
```
kdaily: `raise NotImplementedError`

genie/validate.py
kdaily: Maybe just remove the explicit 'this is valid' instead?

genie/validate.py
```diff
@@ -1,125 +1,160 @@
...
+                        Default is False.
+
+    Returns:
+        message - errors and warnings
+        valid - Boolean value of validation status
+    """
+    if filetype is None:
+        filetype = \
+            determine_filetype(syn, filepathlist, center, raise_error=True)
+    else:
```
kdaily: remove this else.

genie/validate.py
```diff
@@ -1,125 +1,160 @@
...
-            raise ValueError("Your filename is incorrect!\n%s\nPlease change your filename before you run the validator again."  % e)
-    total_error, warning = PROCESS_FILES[fileType](syn, center, threads).validate(filePathList=filePath, oncotreeLink=oncotree_url, testing=testing, noSymbolCheck=noSymbolCheck)
+            filetype = validator.validateFilename(filepathlist)
+        except AssertionError:
+            continue
+        # If valid filename, return file type.
+        if filetype is not None:
+            break
+    if filetype is None and raise_error:
+        raise ValueError(
```
kdaily: move error out (use error or sentinel (`None`) checking, but not both)

genie/input_to_database.py
```diff
@@ -297,13 +242,26 @@ def validatefile(fileinfo,
...
     error_list = check_file_status['error_list']
-    filetype = get_filetype(syn, filepaths, center)
-    if check_file_status['to_validate']:
-
-        valid, message = _check_valid(
-            syn, filepaths, center, filetype, filenames,
-            oncotree_link, threads, testing)
+    # Need to figure out to how to remove this
 
+    filetype = validate.determine_filetype(
```
kdaily: catch error on this if you can't determine file type

kdaily: can remove this if you get `fileType` inside the `check_file_status` dictionary.

genie/validate.py
```diff
@@ -1,125 +1,160 @@
...
-                syn.get(args.uploadToSynapse)
-            except synapseclient.exceptions.SynapseHTTPError as e:
-                raise ValueError("Provided Synapse id must be your input folder Synapse id or a Synapse Id of a folder inside your input directory")
+    valid, message, filetype = validate_single_file(
+        syn, args.filepath, args.center, args.filetype,
+        args.oncotreelink, args.testing, args.nosymbol_check)
 
-    message = validate(syn, args.fileType, args.file, args.center, args.thread, args.oncotreeLink, args.offline, args.uploadToSynapse, args.testing, args.noSymbolCheck)
\ No newline at end of file
+    if args.parentid is not None:
```
kdaily: move this to argparser validation.

kdaily: including the below checking the parent ID

kdaily (changes requested):
I'll rubber stamp this later.

thomasyu888:
Need to add tests for performing validate...

genie/input_to_database.py
thomasyu888: I don't actually track filetype in the table.

kdaily (approved):

PR #137, opened 2019-06-05
Urgent: There are other files with 'clinical' in the filename

thomasyu888:
Did not account for nonGENIE clinical files, so narrowed down the part of the clinical file name I was looking at to detect duplication.


thomasyu888:
Improvement in #138, with tests.  Will be pulled in in the future so that tests will happen prior to pulling in processing PRs

kkdang (approved):

Issue #135, opened 2019-06-03
get all entity information in rename_file


thomasyu888:
Could potentially get all the information of the file entity right here to avoid the syn.get rest call later which doesn't actually download the file

kdaily:
Think this is done.

kdaily:
Rather, it's part of #145.

thomasyu888:
Will close once #145 is merged in

PR #138, opened 2019-06-05
Refactor input_to_db -> Abstraction of getting duplication error

thomasyu888:
* Tests
* Split getting duplication error to two functions
* Lint

### DIFF
--- a/genie/input_to_database.py
+++ b/genie/input_to_database.py
@@ -464,6 +464,69 @@ def create_and_archive_maf_database(syn, database_synid_mappingdf):
     return(database_synid_mappingdf)
 
 
+def email_duplication_error(syn, duplicated_filesdf):
+    '''
+    Sends an email if there is a duplication error
+
+    Args:
+        syn: Synapse object
+        duplicated_filesdf: dataframe with 'id', 'name' column
+    '''
+    if not duplicated_filesdf.empty:
+        incorrect_files = [
+            name for synId, name in zip(duplicated_filesdf['id'],
+                                        duplicated_filesdf['name'])]
+        incorrect_filenames = ", ".join(incorrect_files)
+        incorrect_ent = syn.get(duplicated_filesdf['id'].iloc[0])
+        send_to_users = set([incorrect_ent.modifiedBy,
+                             incorrect_ent.createdBy])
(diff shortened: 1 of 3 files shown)


genie/input_to_database.py
```diff
@@ -464,6 +465,69 @@ def create_and_archive_maf_database(syn, database_synid_mappingdf):
...
+    clinical_files = validation_statusdf.query(
+        'name.str.startswith("data_clinical_supp")')
+    if len(clinical_files) > 2:
+        duplicated_filesdf = duplicated_filesdf.append(clinical_files)
+    duplicated_filesdf.drop_duplicates("id", inplace=True)
+    logger.info("THERE ARE {} DUPLICATED FILES".format(
+        len(duplicated_filesdf)))
+    duplicated_filesdf['errors'] = duplicated_error_message
+    # Send an email if there are any duplicated files
+    email_duplication_error(syn, duplicated_filesdf)
```
kdaily: take this out!

genie/input_to_database.py
```diff
@@ -464,6 +465,69 @@ def create_and_archive_maf_database(syn, database_synid_mappingdf):
...
+        error_email = (
+            "Dear %s,\n\n"
+            "Your files (%s) are duplicated!  FILES SHOULD BE UPLOADED AS "
+            "NEW VERSIONS AND THE ENTIRE DATASET SHOULD BE "
+            "UPLOADED EVERYTIME".format(usernames, incorrect_filenames))
+        syn.sendMessage(
+            list(send_to_users), "GENIE Validation Error", error_email)
+
+
+def get_duplicated_files(syn, validation_statusdf, duplicated_error_message):
```
kdaily, line 525: can you do this way earlier, prior to file validation?

kdaily (changes requested):

genie/input_to_database.py
thomasyu888: Harder to move out than I thought...  The issue is the duplication errors also get saved the invalid status table which has information that is extracted during the validation call (md5, modifiedOn).

This reordering would be possible by building up the information of the entity upon my first rest call. #135.

kdaily (approved):

Issue #124, opened 2019-05-22
define structure for 'fileinfo' rows


kdaily:
```
fileinfo: A row passed in as a Series through the apply function
          in pandas
```

What columns does a row have?

Could this be a class?

thomasyu888:
Removed as of #145

PR #145, opened 2019-06-19
Use entity objects

kdaily:


genie/input_to_database.py
```diff
@@ -67,40 +69,36 @@ def get_center_input_files(syn, synid, center, process="main"):
...
-
-
-def check_existing_file_status(validation_statusdf, error_trackerdf,
-                               entities, input_filenames):
+                clinicalpair.append(ent)
+                continue
+
+            prepared_center_file_list.append([rename_file(syn, ent)])
+    
+    if clinicalpair:
```
thomasyu888: This should be if len(clinicalpair) == 2.  Because if someone uploads just a patient file, it should error out.

kdaily: This function does not make assumptions about why something is bad. it only collects the information. There needs to be a function that 'validates' (for lack of a better word) that the submission package is correct. That logic should be elsewhere.

kdaily: What COULD happen here is that instead of returning file lists, you return an instance of a `FileTypeFormat`, which internally use entities instead of file paths.

tests/unit/test_input_to_database.py
```diff
@@ -421,62 +445,66 @@ def test_invalid_validatefile():
...
-    entity = synapseclient.Entity(id='syn2345', md5='44444')
+    entity = synapseclient.Entity(name="data_clinical_supp_SAGE.txt",
+                                  id='syn1234', md5='44444',
+                                  path='/path/to/data_clinical_supp_SAGE.txt')
     entity['modifiedOn'] = '2019-03-24T12:00:00.Z'
+    # This modifiedOn translates to: 1553428800000
     entity.modifiedBy = '333'
-    entity.createdBy = '333'
-    # entities = [entity]
+    entity.createdBy = '444'
```
thomasyu888:
```suggestion
    entity.createdBy = '333'
```

tests/unit/test_input_to_database.py
```diff
@@ -421,62 +445,66 @@ def test_invalid_validatefile():
...
-              1553428800000,
-              'clinical']],
-            [['syn1234', 'invalid', 'data_clinical_supp_SAGE.txt']])
-        assert foo == expected_validate_results
+            syn, [entity.name], center)
+        patch_get_staterror_list.assert_called_once_with(
+            syn, valid, message, filetype,
+            entities, [1553428800000])
+        patch_send_email.assert_called_once_with(
+            syn, [entity.name], message, [entity.modifiedBy, entity.createdBy])
```
thomasyu888:
```suggestion
            syn, [entity.name], message, [entity.modifiedBy])
```

genie/input_to_database.py
```diff
@@ -291,19 +276,21 @@ def validatefile(fileinfo,
     file_users = list(set([entities[0].modifiedBy, entities[0].createdBy]))
```
thomasyu888:
```suggestion
    file_users = [entities[0].modifiedBy, entities[0].createdBy]
```

thomasyu888:
#135 #124

PR #147, opened 2019-06-20
use the _check_valid function

kdaily:
You have this function written, but it wasn't being used. I updated the signature and used it.


thomasyu888:
Should add tests to make sure this function is called.

genie/input_to_database.py
```diff
@@ -299,21 +298,9 @@ def validatefile(fileinfo,
...
-                testing=testing)
-            logger.info("VALIDATION COMPLETE")
-        except ValueError as e:
-            # Specify this as None for the single case where filename
-            # validation fails
-            filetype = None
-            message = str(e)
-            logger.error(message)
-            valid = False
+        valid, message, filetype = _check_valid(syn, filepaths, center,
```
kdaily: Wrong order of returned parameters.

```suggestion
        valid, filetype, message = _check_valid(syn, filepaths, center, 
```

PR #155, opened 2019-06-25
use argparser to manage exclusive options

kdaily:


genie/__main__.py
```diff
@@ -65,6 +65,13 @@ def build_parser():
         help='Contributing Centers')
 
     parser_validate.add_argument(
+        "--oncotreelink",
+        type=str,
+        help="Link to oncotree code")
+
+    group = parser.add_mutually_exclusive_group()
```
thomasyu888: Wait but does this add the parameter to `parser_validate`? It has to be a parameter in `genie validate -h` @kdaily

kdaily: Fixed by 7c04fa4.

thomasyu888 (approved):

PR #159, opened 2019-06-27
Refactor validate command line

kdaily:
Not really refactor, but format, lint, and make `perform_validate` that is used by the CLI a private function so it's not exposed to end user.


thomasyu888:
Why not use `main()`?

thomasyu888 (approved):

thomasyu888 (approved):

PR #160, opened 2019-06-27
get rid of try/catch to find one specific error.

kdaily:
Just append errors to be handled by the warning and error log handler.

This is a starting point - I think that the `FileTypeFormat` could return `(valid, message, warning)` to simplify further.

### DIFF
--- a/genie/input_to_database.py
+++ b/genie/input_to_database.py
@@ -166,21 +166,16 @@ def _check_valid(syn, filepaths, center, filetype, filenames,
     Function to validate a file
     '''
     # If no filetype set, means the file was named incorrectly
-    try:
-        valid, message, filetype = validate.validate_single_file(
-            syn,
-            filepaths,
-            center,
-            oncotreelink=oncotree_link,
-            testing=testing)
-        logger.info("VALIDATION COMPLETE")
-    except ValueError as e:
-        # Specify this as None for the single case where filename
-        # validation fails
-        filetype = None
-        message = str(e)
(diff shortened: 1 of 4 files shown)


thomasyu888:
Also, there are a couple classes that throw ValueErrors in `_get_dataframe`.  These errors would have to be caught as well and appeneded.

genie/input_to_database.py
```diff
@@ -166,21 +166,16 @@ def _check_valid(syn, filepaths, center, filetype, filenames,
...
-        valid = False
+    valid, message, filetype = validate.validate_single_file(
+        syn,
+        filepaths,
+        center,
+        oncotreelink=oncotree_link,
+        testing=testing)
+    logger.info("VALIDATION COMPLETE")
+
+    valid, message = determine_validity_and_log(total_error, warning)
```
thomasyu888: This is unnecessary as it is already done in `validate_single_file`

thomasyu888 (changes requested):

thomasyu888:
Will also have to change tests in `test_input_to_database.py`: `test_filetype_validate_single_file`, and `test_wrongfiletype_validate_single_file` and `ValueErrors` are no longer thrown.  Will have to check for `total_error`

genie/input_to_database.py
kdaily: removed function as it isn't doing anything extra for validate single file

thomasyu888 (approved):

PR #176, opened 2019-08-29
Make validate class

kdaily:
This PR converts the validator module into a validator class. This is really a validation helper class, as it's job is to determine which registered validator to use.

~It needs more initialization, which I will work on next. This is the minimal change to implement the class as is to get all of the existing tests to pass.~

It now instantiates with a center and format registry dictionary as well.

I should convert the registry dictionary to an immutable object to prevent changes.

### DIFF
--- a/genie/input_to_database.py
+++ b/genie/input_to_database.py
@@ -259,12 +259,10 @@ def validatefile(syn, entities, validation_statusdf, error_trackerdf,
     # Need to figure out to how to remove this
     # This must pass in filenames, because filetype is determined by entity name
     # Not by actual path of file
-    validator = validate.Validator(syn=syn, center=center)
-    filetype = validator.determine_filetype(filenames)
+    validator = validate.Validator(syn=syn, center=center, filepathlist=filepaths)
+    filetype = validator.file_type
     if check_file_status['to_validate']:
         valid, message, filetype = validator.validate_single_file(
-            filepaths,
-            filetype=filetype,
             oncotreelink=oncotree_link,
             testing=testing)
         logger.info("VALIDATION COMPLETE")
--- a/genie/input_to_database.py
+++ b/genie/input_to_database.py
@@ -259,7 +259,7 @@ def validatefile(syn, entities, validation_statusdf, error_trackerdf,
     # Need to figure out to how to remove this
     # This must pass in filenames, because filetype is determined by entity name
     # Not by actual path of file
-    validator = validate.Validator(syn=syn, center=center, filepathlist=filepaths)
+    validator = validate.GenieValidationHelper(syn=syn, center=center, filepathlist=filepaths)
     filetype = validator.file_type
     if check_file_status['to_validate']:
         valid, message, filetype = validator.validate_single_file(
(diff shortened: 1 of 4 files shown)


thomasyu888 (approved):

genie/input_to_database.py
```diff
@@ -259,13 +259,10 @@ def validatefile(syn, entities, validation_statusdf, error_trackerdf,
     # Need to figure out to how to remove this
     # This must pass in filenames, because filetype is determined by entity name
     # Not by actual path of file
-    filetype = validate.determine_filetype(syn, filenames, center)
+    validator = validate.Validator(syn=syn, center=center, filepathlist=filepaths)
```
kdaily: this should really be explicitly parameterized with the format registry dict instead of using the default.

genie/validate.py
```diff
@@ -12,31 +12,108 @@
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.INFO)
 
+class Validator(object):
```
kdaily: this should be an `ABCMeta` class.

genie/validate.py
```diff
@@ -12,31 +12,108 @@
...
+
+        Returns:
+            str: File type of input files.  None if no filetype found
+        '''
+        filetype = None
+        # Loop through file formats
+        for file_format in self._format_registry:
+            validator = self._format_registry[file_format](self._synapse_client, self.center)
+            try:
+                filetype = validator.validateFilename(self.filepathlist)
```
kdaily, line 46: `example_file_format.validateFilename` should be something like `example_file_format.validate_file_type`.

genie/validate.py
```diff
@@ -12,31 +12,108 @@
...
+            except AssertionError:
+                continue
+            # If valid filename, return file type.
+            if filetype is not None:
+                break
+        return(filetype)
+
+    def validate_single_file(self, filetype=None,
```
kdaily, line 54: if this was an `ABCMeta` then `oncotreelink`, etc go away.

genie/validate.py
kdaily: `ValidateHelper`? `ValidatorWrapper`?

genie/validate.py
```diff
@@ -209,8 +241,8 @@ def _perform_validate(syn, args):
     args.oncotreelink = _get_oncotreelink(syn, databasetosynid_mappingdf,
                                           oncotreelink=args.oncotreelink)
 
-    valid, message, filetype = validate_single_file(
-        syn, args.filepath, args.center, args.filetype,
+    validator = Validator(syn=syn, center=args.center, filepathlist=args.filepath)
```
kdaily: this function should go to the command line file, not here

genie/validate.py
kdaily: Interim solution is to make this like an `ABCMeta` class, and make a separate `GenieValidatorHelper` that inherits from it.

genie/validate.py
kdaily: This is a note only - requires bigger refactor. #185

genie/validate.py
kdaily: not doing this now because requires reconfiguring test of the function.

genie/input_to_database.py
kdaily: done

PR #184, opened 2019-09-06
Move tests out of unit folder

thomasyu888:
Many of these tests aren't actually unit tests, therefore should not belong in a `unit` folder.


kdaily (approved):

kdaily:
Did you want this to go directly to `develop`?

thomasyu888:
Yup. Thanks! Actually, we should've probably been putting things that were unrelated to `database_to_staging` to `develop`, but its fine.  Its just a common branch that we could work on.

PR #194, opened 2019-09-13
Add db to staging script to setup

thomasyu888:


kdaily (approved):
Looks good. Tom is filing a separate issue to add these scripts to a common cli
#204

PR #195, opened 2019-09-17
Format registry as list

kdaily:


genie/validate.py
```diff
@@ -207,6 +211,29 @@ def _upload_to_synapse(syn, filepaths, valid, parentid=None):
...
+    Returns:
+        A list of classes that are in the named packages and subclasses of example_filetype_format.FileTypeFormat.
+    """
+
+    file_format_list = []
+    for package_name in package_names:
+        importlib.import_module(package_name)
+
+    for cls in config.get_subclasses(example_filetype_format.FileTypeFormat):
+        logger.debug(f"checking {cls}.")
```
thomasyu888:
```suggestion
        logger.debug("checking {}.".format(cls))
```

thomasyu888: Support python 3.5 for now.

genie/validate.py
```diff
@@ -228,8 +255,12 @@ def _perform_validate(syn, args):
     args.oncotree_link = _get_oncotreelink(syn, databasetosynid_mappingdf,
                                            oncotree_link=args.oncotree_link)
 
+    format_registry = collect_format_types(args.format_registry_packages)
+    logger.debug(f"Using {format_registry} file formats.")
```
thomasyu888:
```suggestion
    logger.debug("Using {} file formats.".format(format_registry))
```

genie/validate.py
```diff
@@ -228,8 +255,12 @@ def _perform_validate(syn, args):
     args.oncotree_link = _get_oncotreelink(syn, databasetosynid_mappingdf,
                                            oncotree_link=args.oncotree_link)
 
+    format_registry = collect_format_types(args.format_registry_packages)
```
thomasyu888, line 258: Need to add `args.format_registry_packages` to `__main__.py`

thomasyu888 (changes requested):

thomasyu888 (commented):
Would determining which file formats to validate and process come after collecting all file formats.  That is if we were to add a parameter like include/exclude types, where should the filtering happen?

kdaily:
I think so. There's no way to a priori know which files to get via `get_center_input_files`. I would:

1. Get all center input files
2. Determine their file type
3. Filter based on include/exclude

thomasyu888:
The issue with that is that there are tons of vcf files. So I don't want to necessarily download them.

kdaily:
Agree - but relying on the file name (without using the `determineFiletype` function(s)) to determine the type is not going to work. Is it the case currently that you can use the entity name, NOT the file handle file name, to determine the file type?

thomasyu888:
Ah I see what you're saying. Yes we can infer the file type with the entity name.

kdaily:
`get_center_input_files` could do that, with a `download` parameter passed on to the `syn.get` call.

However, I was thinking that to simplify much of the other code, that the entity itself can be used to carry updated information as it moves through validation and processing (it's mutable). If the entity is retrieved multiple times this would be difficult to manage - but a wrapper object could be used.

1. `syn.get(..., download=False)` -> entity
2. `determineFiletype` updates the entity with the fileType
3. filter the entity based on the file type
4. download the entity for validation or processing and update any info added.

Proposed wrapper could be something like `GenieEntity(entity=entity, info=dict(fileType='vcf', ...))`. Provide a replace entity function that checks to be sure the same entity ID/version is supplied?

thomasyu888 (approved):

PR #199, opened 2019-09-19
add filetype to error table and fix tests.

kdaily:
This change fixes the rest of the file types for the error table, and fixes tests for it.

It also updates the functions that are responsible for manipulating the status and error tables to always use the query result object instead of passing the dataframe separately.

### DIFF
--- a/genie/input_to_database.py
+++ b/genie/input_to_database.py
@@ -214,7 +214,7 @@ def _get_status_and_error_list(valid, message, filetype, entities):
              ent.name, entity_date_to_timestamp(ent.properties.modifiedOn), filetype]
             for ent in entities]
         invalid_errors_list = [
-            [ent.id, message, ent.name]
+            [ent.id, message, ent.name, filetype]
             for ent in entities]
     return(input_status_list, invalid_errors_list)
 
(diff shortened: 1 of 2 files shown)


genie/input_to_database.py
```diff
@@ -250,7 +253,7 @@ def validatefile(syn, entities, validation_statusdf, error_trackerdf,
     file_users = [entities[0].modifiedBy, entities[0].createdBy]
 
     check_file_status = check_existing_file_status(
-        validation_statusdf, error_trackerdf, entities)
+        validation_status_table, error_tracker_table, entities)
```
thomasyu888, line 256: Any reason you switch to using `table` instead of `df`?

kdaily: I noted in the issue description that some fxns use one and some use the other. I made it standard.

genie/input_to_database.py
```diff
@@ -602,27 +607,26 @@ def validation(syn, center, process,
...
-        error_trackerdf = error_tracker_table.asDataFrame()
 
         input_valid_statuses = []
         invalid_errors = []
 
         for ents in center_files:
             status, errors = validatefile(syn, ents,
-                                          validation_statusdf,
-                                          error_trackerdf,
+                                          validation_status_table,
```
thomasyu888, line 628: Need to edit the doc string in `validatefile`. Currently it says it takes in the dataframe.

thomasyu888 (approved):

thomasyu888 (approved):

PR #203, opened 2019-09-23
Create gene panel should happen outside of the bed class

thomasyu888:


thomasyu888:
Begin to strip out create gene panel.  Incomplete.

kdaily (approved):

PR #211, opened 2019-11-01
Temp file for update

kdaily:
Use a temporary file for updating database. This allows multiple processes to operate at the same time.


kdaily:
Oops. Pulled from wrong remote.

PR #218, opened 2019-11-19
Remove testing logic

kdaily:
This PR removes logic in the file format, input, and validation functions to get configuration for a production or testing/staging Synapse project. It instead relies on receiving as input a Synapse Project ID. This Synapse project must be configured according to GENIE infrastructure standards, including having a database mapping table, and having an annotation at the Project level where `dbMapping` points to the Synapse ID of that table.

The tests were fixed for `input_to_database`, `validate`, and `process_functions` (except for one `process_functions` test that is failing that I cannot figure out).

The tests for individual file formats largely fail because they still take a 'test' parameter, which was removed from the `example_file_format.FileFormat` class.


kdaily:
Closing due to accidentally pulling in changes from another fork.

Issue #119, opened 2019-05-21
Use thread argparse in input_to_database


thomasyu888:
This parameter isn't currently used but should be.  

Currently the number of threads used for `vcf` and `maf` processing is hard coded

thomasyu888:
Won't do.

PR #219, opened 2019-11-20
remove all references to threads

kdaily:
Threading is unused, and the references are just littering the function definitions. Parallelization in the general case should be handled at the batch job level (e.g., AWS Batch). For the few cases where multiple files are consumed at once, they can locally configure threading options.


kdaily:
Using threading appropriately would cover #119, but this should not be in the `input_to_database` level.

kdaily:
Argh. This is broken because I merged the same branch into my fork but had to update due to merge conflicts. I will try to revert 585d1d8 to fix this.

kdaily:
Closing, and re-opening another PR without extraneous changes.

PR #220, opened 2019-11-21
Separate maf database creation from cli

kdaily:
Creating a MAF database should happen separately - it has no dependence on the rest of the arguments or processing. This PR moves the code to create a MAF database to a new command line tool and removes the functionality from `input_to_database`.


thomasyu888:
This is definitely a good move.  This was merely for 'convenience' as I would always process maf or vcf first.  There should really be a separate batch job that is: 'create maf db' and that gets run before anything else when new processing starts.

I like this revision, but will have to add/revise batch jobs

kdaily:
Thanks! And yes it will require changes to the infrastructure. Can I merge this and open a separate issue (here and/or in the Genie AWS infra repo) to fix that?

thomasyu888 (approved):

bin/create_maf_database.py
```diff
@@ -0,0 +1,32 @@
...
+from genie import input_to_database
+from genie import process_functions
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+def main(db_mapping_id, pemfile):
+
+    syn = process_functions.synLogin(pemfile, debug=debug)
```
thomasyu888, line 14: Missing debug parameter

bin/create_maf_database.py
```diff
@@ -0,0 +1,32 @@
...
+    query = 'SELECT * FROM {}'.format(db_mapping_id)
+    databaseToSynIdMapping = syn.tableQuery(query)
+    databaseToSynIdMappingDf = databaseToSynIdMapping.asDataFrame()
+
+    databaseToSynIdMappingDf = \
+            input_to_database.create_and_archive_maf_database(syn, databaseToSynIdMappingDf)
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--db_mapping_id', type=str,
```
thomasyu888, line 25: Will this just be projectid at some point since we will be adding this as an annotation?

thomasyu888 (changes requested):

thomasyu888:
@kdaily, must also add to setup.py

bin/create_maf_database.py
thomasyu888: Will update this parameter once this is fixed: #222

thomasyu888:
The true functionality of this function is
* Create a new table for a specific file format
* Replace the database mapping table with the newly created table.

PR #224, opened 2019-11-22
remove gzip dependency, seems unused

kdaily:


thomasyu888:
It was used here: https://github.com/Sage-Bionetworks/Genie/blob/565b0d0371c9ed44ce3927f2d223772ea8fa639d/genie/bed.py#L76. 

I'm not sure it's necessary now, since I mock the step where the large file is downloaded in bed tests.

kdaily:
I searched for `gunzip` in the codebase and it didn't occur!

thomasyu888:
It's there, but it's not used in the tests any more. I think.... Since Travis is meant to run tests, I think it's ok to remove this dependency.

thomasyu888 (approved):

PR #229, opened 2020-01-12
DEV: No duplicated variants in vcf / maf

thomasyu888:
* Duplicated variants flagged as warnings
* duplicated variants removed prior to upload into database

### DIFF
--- a/genie/maf.py
+++ b/genie/maf.py
@@ -214,14 +214,30 @@ class maf(FileTypeFormat):
 
         total_error = ""
         warning = ""
-        # CHECK: First column must be in the first_header list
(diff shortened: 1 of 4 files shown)


genie/maf.py
```diff
@@ -293,6 +293,10 @@ def _validate(self, mutationDF):
                     "CHROMOSOME column cannot have any values that "
                     "start with 'chr' or any 'WT' values.\n")
 
+        # No duplicated values
+        if mutationDF.duplicated().any():
```
kkdang: Add in columns to check duplicates by, dont use all the columns

kkdang (approved):

PR #230, opened 2020-01-12
DEV: Email warnings

thomasyu888:
* Contemplate adding another status
VALIDATED, INVALID, PASS... `CHECKON`?


thomasyu888:
Closing as this is not the right way to email warnings to centers.

PR #235, opened 2020-01-31
Pass in entity list into genie validator to obtain filetype

thomasyu888:
Instead of passing in the filepath, pass in a list of entities.  This is important because the `entity.name` is used to determine the filetype.


kdaily (commented):
This should probably merge into develop, but I think it looks good!

kdaily:
@thomasyu888 and I discussed and this is a patch to something broken, so it should go into master.

kkdang (approved):

PR #242, opened 2020-02-19
Make sure the center is part of the centers being processed

thomasyu888:
Bed file processing was running into `center not found` index issue, so make sure that the `center` is being processed.


thomasyu888:
This is not the issue. Closing

PR #246, opened 2020-02-23
Fix ethnicity race dashboard

thomasyu888:
* ETHNICITY and RACE dashboards were not showing those with 100% Unknown.


xindiguo (approved):
Looks good

PR #266, opened 2020-03-26
Fix filetype flag

thomasyu888:
#241


xindiguo (approved):
Looks good!

thomasyu888:
Some sites require this flag because their pipeline doesn't output files in the correct filename.  They use the entity name on synapse to upload correctly. Will pull in.

Issue #283, opened 2020-04-16
PHI cutoff should be greater or equal to 90 years in days


xindiguo:
The interval cutoff should be >= math.floor(90*365.24)

https://github.com/Sage-Bionetworks/Genie/blob/cfa33533026cb09d7d56616ebc516942a9415bcb/genie/database_to_staging.py#L105

thomasyu888:
The BPC redaction policy should match the GENIE one.  redact 89 years by days.

PR #290, opened 2020-04-20
Update redaction

thomasyu888:
#283


thomasyu888:
Decided that we will be redacting patients over 89 **by days** and not by years.  Therefore, no need to change the redaction policy.

PR #291, opened 2020-04-24
Format SEQ_PIPELINE_ID and add test

thomasyu888:
Make sure SEQ_PIPELINE_ID is in the same format as SEQ_ASSAY_ID:

* Uppercase
* Change underscore to dash


xindiguo (approved):
Approved.

PR #299, opened 2020-05-11
Update database mapping synapse id so that correct table is updated

### LINKED ISSUE #274: Fix createmafdb
Incorrect synapse id
```
    syn.store(synapseclient.Table("syn12094210", vcf2maf_mappingdf))
```

thomasyu888:
When creating maf database, the database mapping table is updated.


genie/__version__.py
```diff
@@ -1 +1 @@
-__version__ = "9.0.0"
+__version__ = "10.0.0-dev"
```
```suggestion
__version__ = "9.0.0"
```

xindiguo (approved):
Good to go

PR #300, opened 2020-05-11
Fix dashboardTemplate.Rmd panel issue

thomasyu888:
* Not all panels will be present in the mutated_gene table.
* Table of contents renders weird in synapse html preview


xindiguo (approved):
Looks good.

PR #313, opened 2020-05-20
frequently mutated genes dashboard table should only show non-synonymous variants

### LINKED ISSUE #310: Adjust frequently mutated genes to only have non-synonymous variants
Any variants that are: Silent, Intron, IGR, 3'UTR, 5'UTR, 3'Flank and 5'Flank and have blank HGVSp_Short is filtered out with the exception of TERT promoters?

thomasyu888:
The frequently mutated genes from the maf file should only be from non-synonymous variants.
fixes #310


genie/__version__.py
```diff
@@ -1 +1 @@
-__version__ = "10.0.0"
+__version__ = "11.0.0-dev"
```
```suggestion
__version__ = "10.0.0"
```

kkdang (approved):

PR #325, opened 2020-06-04
Add test for removeStringFloat()

### LINKED ISSUE #323: Add tests for removeStringFloat
https://github.com/Sage-Bionetworks/Genie/blob/d39451655ec3632df6002c1d73b17dacba2a8720/genie/process_functions.py#L306-L318

karawoo:
Closes #323. Let's hold off on reviewing until we can do it together in the code review meeting tomorrow :)

One question I had is whether string representations of TSV files always end in a newline? I had originally written the last test case as `"'a'\t'b'\n1.0\t2.0"`, but then the `2.0` doesn't get changed. If it's feasible that the function might receive data like this and the 2.0 should get changed to 2, then that would require some adjustments to the code.


thomasyu888:
Thanks for the contribution,  I have prepared a review that we can go through during the meeting

tests/test_process_functions.py
```diff
@@ -17,6 +17,17 @@
 DATABASE_DF.index = ['1_3', '2_3', '3_5']
 ENTITY = synapseclient.Project("foo", annotations={"dbMapping": ["syn1234"]})
 
+def test_removeStringFloat():
```
thomasyu888:
```

@pytest.mark.parametrize("input_str,output", [
        ("1.0\t", "1\t"),
        ("1.0\n", "1\n"),
        ("1.5\t", "1.5\t"),
        ("1\t", "1\t"),
        ("0\t", "0\t"),
        ("'a'\t'b'\n1.0\t2.0\n", "'a'\t'b'\n1\t2\n"),
    ])
def test_removeStringFloat(input_str, output):
    """Remove string float - will always assume that there is a \n
    at the end.  This is because if a value was 2.01, we dont want to
    remove the .0 from this."""
    assert genie.process_functions.removeStringFloat(input_str) == output
```

karawoo: Thanks @thomasyu888! I've updated with your suggestions

tests/test_process_functions.py
```diff
@@ -17,6 +17,21 @@
...
 
+@pytest.mark.parametrize("input_str,output", [
+        ("1.0\t", "1\t"),
+        ("1.0\n", "1\n"),
+        ("1.5\t", "1.5\t"),
+        ("1\t", "1\t"),
+        ("0\t", "0\t"),
+        ("'a'\t'b'\n1.0\t2.0\n", "'a'\t'b'\n1\t2\n"),
+    ])
+
```
thomasyu888:
```suggestion
```

tests/test_process_functions.py
```diff
@@ -17,6 +17,20 @@
...
+        ("1\t", "1\t"),
+        ("0\t", "0\t"),
+        ("'a'\t'b'\n1.0\t2.0\n", "'a'\t'b'\n1\t2\n"),
+    ])
+def test_removeStringFloat(input_str, output):
+    """Remove string float - will always assume that there is a \n
+    at the end.  This is because if a value was 2.01, we dont want to
+    remove the .0 from this."""
+    assert genie.process_functions.removeStringFloat(input_str) == output
+
```
thomasyu888, line 33:
```suggestion


```

thomasyu888 (approved):

PR #326, opened 2020-06-08
v10

thomasyu888:
* Move `consortium_to_public` cli out of genie package into `bin`
* Use `unittest.mock`
* Remove support for `synapseclient < 2`


genie/__version__.py
```diff
@@ -1 +1 @@
-__version__ = "10.0.0"
+__version__ = "11.0.0-dev"
```
```suggestion
__version__ = "10.0.0"
```

PR #354, opened 2020-09-15
v12

thomasyu888:
* Split GENIE file format registry from GENIE package
* Added `create-case-lists` cli
* use assay information to create case list files instead of gene matrix file
* Improve clinical attribute mapping during input to database step
* Support pandas>=1.1
* Added create new database functions
* Added `get-invalid-errors` cli function


genie/__main__.py
```diff
@@ -34,6 +33,15 @@ def synapse_login(username=None, password=None):
     return syn
 
 
+def perform_create_case_list(syn, args):
+    """CLI to create case lists given a clinical and assay information file"""
+    create_case_lists.main(args.clinical_file_name,
+                           args.assay_info_file_name,
+                           args.output_dir,
+                           args.study_id)
+
```
thomasyu888, line 42:
```suggestion
```

genie/__version__.py
```diff
@@ -1 +1 @@
-__version__ = "11.0.0"
+__version__ = "12.0.0-dev"
```
```suggestion
__version__ = "12.0.0"
```

PR #362, opened 2020-10-13
clean up v12

thomasyu888:
Clean up repo: Remove scripts that weren't used, create `template` and `R` directory


genie/__version__.py
```diff
@@ -1 +1 @@
-__version__ = "12.0.0"
+__version__ = "12.1.0-dev"
```
```suggestion
__version__ = "12.1.0"
```

PR #366, opened 2020-11-17
12.2.0 release

thomasyu888:
* Add provenance to tables and release files
* Update synapseclient


genie/__version__.py
```diff
@@ -1 +1 @@
-__version__ = "12.1.0"
+__version__ = "12.2.0-dev"
```
```suggestion
__version__ = "12.2.0"
```

PR #415, opened 2022-01-03
Check for missing seq assay id in the assay information file

thomasyu888:
The code pulls from the sample clinical database and checks the assay information file against it to see if there are any missing SEQ_ASSAY_IDs.   There are a couple unfortunate parts about this not being a cross file check but a check against the database:

1. The clinical database could not be updated with the newest clinical file
2. If the assay information isn't changed, validation won't be triggered for it. This means, we will have to manually trigger its validation.

This is the error message:
```
ERROR:genie.validate:Assay_information.yaml: You are missing SEQ_ASSAY_IDs: VICC-01-D2, VICC-01-T4B, VICC-01-T6B, VICC-01-DX1
```


genie_registry/assay.py
```diff
@@ -135,6 +137,26 @@ def _validate(self, assay_info_df):
                 total_error += \
                     ("Assay_information.yaml: Please make sure your all your "
```
hhunterzinck: extra 'your' in error message

genie_registry/assay.py
```diff
@@ -144,18 +144,20 @@ def _validate(self, assay_info_df, project_id):
...
+                f"select distinct(SEQ_ASSAY_ID) as seq from {sample_synid} "
+                f"where CENTER = '{self.center}'"
             )
             uniq_seq_df = uniq_seq_table.asDataFrame()
             # These are all the SEQ_ASSAY_IDs that are in the clinical database
             # but not in the assay_information file
-            missing_seqs = uniq_seq_df['uniq_seq'][
-                ~uniq_seq_df['uniq_seq'].isin(all_seq_assays)
+            missing_seqs = uniq_seq_df['seq'][
+                ~uniq_seq_df['seq'].isin(all_seq_assays)
```
hhunterzinck, line 154: Are the elements of all_seq_assays guaranteed to be an exact match (capitalization, hyphen versus underscore) to values in the sample file?

hhunterzinck (approved):

genie_registry/assay.py
thomasyu888: It is up to the sites themselves to format their SEQ_ASSAY_IDs correctly, so if their seq assay ids don't match whats in the clinical file, then it should probably fail.

genie_registry/assay.py
```diff
@@ -135,6 +137,28 @@ def _validate(self, assay_info_df):
                 total_error += \
                     ("Assay_information.yaml: Please make sure your all your "
```
thomasyu888:
```suggestion
                    ("Assay_information.yaml: Please make sure all your "
```

PR #416, opened 2022-01-03
check for vital status inconsistencies

thomasyu888:
Check for Vital status column inconsistencies.  The error message looks like this:

```
ERROR:genie.validate:Patient: you have inconsistent values in YEAR_CONTACT, INT_CONTACT
ERROR:genie.validate:Patient: you have inconsistent values in YEAR_DEATH, INT_DOD
ERROR:genie.validate:Patient: you have inconsistent values in YEAR_DEATH, DEAD, INT_DOD
```


genie_registry/clinical.py
```diff
@@ -77,6 +77,36 @@ def _check_year(clinicaldf: pd.DataFrame, year_col: int, filename: str,
...
+
+    # cols = ["YEAR_CONTACT", "INT_CONTACT"]
+    # missing_vals = ["Not Collected", "Unknown", "Not Applicable"]
+    # Get index of all rows that have 'missing' values
+    check_inconsistencies = clinicaldf[cols].isin(missing_vals)
+    # unique missing values per column
+    uniq_missing_values = check_inconsistencies.sum(axis=0).unique()
+    if len(uniq_missing_values) > 1:
+        return f"Patient: you have inconsistent values in {', '.join(cols)}\n"
+    return ""
```
hhunterzinck: This implementation will pass a file that has the same number of missing values in each column but not necessarily on the same rows.  For example, the following df will pass even though it has inconsistent data for two rows: 

`clinicaldf = pd.DataFrame([[1, "Unknown"], ["Unknown", 3]], columns = cols)
`

Perhaps instead: 

`n_missing = clinicaldf[cols].isin(missing_vals).sum(axis=1)
if n_missing.between(0, len(cols), inclusive="neither").any():
    print(f"Patient: you have inconsistent values in {', '.join(cols)}\n")
else:
    print("ok")
`

tests/test_clinical.py
```diff
@@ -705,8 +705,27 @@ def test__check_vital_status_consistentency_valid():
...
-        cols=['INT2', "YEAR1", "FOO_3"],
+        cols=['INT_2', "YEAR_1", "FOO_3"],
         missing_vals=["Unknown"]
-    )
\ No newline at end of file
+    )
+    assert error == ""
+
+
+def test__check_vital_status_consistentency_inconsistent():
```
hhunterzinck: Suggestion for another check that should result in an error:

`clinicaldf = pd.DataFrame([[1, "Unknown"], ["Unknown", 3]], columns = cols)`

hhunterzinck (changes requested):
Suggested an alternative implementation to the vital status inconsistencies check that will catch inconsistent values across rows.  Also, suggested additional check to verify the check catches this case.

tests/test_clinical.py
thomasyu888: We should also do something about the ">..." and "<...." values.  For instance if a YEAR_CONTACT column has ">89", the INT_CONTACT should have ">..." as well.

hhunterzinck: agreed

thomasyu888: Ah.. I also realize that the unknown values should probably map to one another.  So if there is a "Unknown" it should technically map to an "Unknown" and not "Not Collected".  For instance, currently this will pass validation...

```
clinicaldf = pd.DataFrame([[1, "Unknown"], [3, "Not Collected"]], columns = cols)
```

hhunterzinck: Did you mean...

`clinicaldf = pd.DataFrame([[1, 3], ["Unknown", "Not Collected"]], columns = cols)` ?

I think it would depend on the definition of the schema.  If all of the above are valid as representations of missing values, then having a mixture is fine even if the actual English word meanings are technically different.

thomasyu888: Ah yes, thats what I meant.  The problem is those string values mean different things.  So if `INT_DOD` is "Not Applicable", `YEAR_DEATH` should also be "Not Applicable".  It shouldn't be "Unknown..

The definition of those values are:

Not Released: Information cannot be released due to site IRB.
Not Collected: Information not collected (ie. PRIMARY_RACE isn't collected at some sites)
Not Applicable: Information cannot be filled out (ie. YEAR_DEATH can't be filled out if the patient is still alive)
Unknown: Patient/Sample information collected, but is unknown.

hhunterzinck: I see.  Then I agree since they have specific meanings, the labels should be consistent.

genie_registry/clinical.py
```diff
@@ -77,6 +77,92 @@ def _check_year(clinicaldf: pd.DataFrame, year_col: int, filename: str,
...
+        clinicaldf.loc[is_str, "DEAD"] == clinicaldf.loc[is_str, "INT_DOD"]
+    )
+    # If dead, int column can't be Not Applicable
+    # If alive, int column can't have values
+    if (any(clinicaldf.loc[is_dead, 'INT_DOD'] == "Not Applicable") or
+            not all(clinicaldf.loc[is_alive, 'INT_DOD'].isin(allowed_str)) or
+            not is_equal):
+        return (
+            "Patient: you have inconsistent values in "
+            "INT_DOD, YEAR_DEATH, DEAD\n"
```
hhunterzinck: Would suggest changing the comment to explicitly indicate the DEAD variable needs to be changed.  Maybe something like:

"Patient Clinical File: DEAD value is inconsistent with INT_DOD for at least one patient"

genie_registry/clinical.py
```diff
@@ -77,6 +77,92 @@ def _check_year(clinicaldf: pd.DataFrame, year_col: int, filename: str,
...
+        # n string values per row
+        n_str = (clinicaldf[cols] == str_val).sum(axis=1)
+        if n_str.between(0, len(cols), inclusive="neither").any():
+            is_inconsistent = True
+
+    # Check that the redacted values are consistent
+    is_redacted_int_89 = clinicaldf[interval_col] == ">32485"
+    is_redacted_year_89 = clinicaldf[year_col] == ">89"
+    is_redacted_int = clinicaldf[interval_col] == "<6570"
+    is_redacted_year = clinicaldf[year_col] == "<18"
```
hhunterzinck, line 157: Would recommend using a regex for the '>' and '<' matching instead of matching on the exact value to be more robust

genie_registry/clinical.py
```diff
@@ -77,6 +77,92 @@ def _check_year(clinicaldf: pd.DataFrame, year_col: int, filename: str,
...
+    if (any(is_redacted_int != is_redacted_year) or
+            any(is_redacted_int_89 != is_redacted_year_89)):
+        is_inconsistent = True
+
+    if is_inconsistent:
+        return f"Patient: you have inconsistent values in {', '.join(cols)}\n"
```
hhunterzinck, line 162: Similar to identifying inconsistencies with the DEAD variable, it would be great if the error message could include the patient IDs for rows with inconsistent values to facilitate identification of these rows.

f"Patient Clinical File: you have inconsistent values in {', '.join(cols)} for the following patients: GENIE-xxx-xxxx, GENIE-xxx-xxxx\n"

hhunterzinck (changes requested):
Checks look logically consistent.  Main suggestion is to increase the utility of the error messages by including patient IDs corresponding to rows with invalid combinations of values in order to make the issue easier for users to identify.

genie_registry/clinical.py
hhunterzinck: Values are locked and will not change.  Other values with < and > are invalid so need to check for exact values

genie_registry/clinical.py
```diff
@@ -77,6 +77,92 @@ def _check_year(clinicaldf: pd.DataFrame, year_col: int, filename: str,
...
+        # This is assuming that interval and year columns start with
+        # INT/YEAR
+        interval_col = col if col.startswith("INT") else interval_col
+        year_col = col if col.startswith("YEAR") else year_col
+        # Return empty string is columns don't exist because this error
+        # is already handled.
+        if not process_functions.checkColExist(clinicaldf, col):
+            return ""
+
+    is_inconsistent = False
```
hhunterzinck: to separate out the two types of inconsistencies, maybe define two flags here:

1. is_missing_inconsistent
2. is_redaction_inconsistent

and set flags in appropriate locations.

genie_registry/clinical.py
```diff
@@ -77,6 +77,92 @@ def _check_year(clinicaldf: pd.DataFrame, year_col: int, filename: str,
...
+    is_redacted_year_89 = clinicaldf[year_col] == ">89"
+    is_redacted_int = clinicaldf[interval_col] == "<6570"
+    is_redacted_year = clinicaldf[year_col] == "<18"
+    if (any(is_redacted_int != is_redacted_year) or
+            any(is_redacted_int_89 != is_redacted_year_89)):
+        is_inconsistent = True
+
+    if is_inconsistent:
+        return f"Patient: you have inconsistent values in {', '.join(cols)}\n"
+    return ""
```
hhunterzinck: if is_redaction_inconsistent and is_missing_inconsistent: "Patient: you have inconsistent redaction and text values in cols"
else if is_redaction_inconsistent: "Patient: you have inconsistent redaction in cols"
else if is_missing_inconsistent: "Patient: you have inconsistent text values in cols"

hhunterzinck (approved):

PR #418, opened 2022-01-06
Release v12.6.0

thomasyu888:
* Add in vital status consistency check
* Remove python 3.6 due to EOL
* Update Dockerfile


genie/__version__.py
```diff
@@ -1 +1 @@
-__version__ = "12.5.0"
+__version__ = "12.6.0-dev"
```
```suggestion
__version__ = "12.6.0"
```

PR #436, opened 2022-02-02
make seq_assay_id check case insensitive

### LINKED ISSUE #435: Make validation of SEQ_ASSAY_IDs case-insensitive
The SEQ_ASSAY_IDs in the assay_information.yaml are being compared to the 'Sample Clinical Database' table (syn7517674), which converts the SEQ_ASSAY_IDs to all uppercase.  If the SEQ_ASSAY_IDs assay_information.yaml contain lowercase characters, they will be erroneously flagged.

hhunterzinck:
Fixes #435

### DIFF
--- a/genie_registry/assay.py
+++ b/genie_registry/assay.py
@@ -141,8 +141,7 @@ class Assayinfo(FileTypeFormat):
         warning = ""
 
         if process_functions.checkColExist(assay_info_df, "SEQ_ASSAY_ID"):
-            all_seq_assays_raw = assay_info_df.SEQ_ASSAY_ID.unique()
-            all_seq_assays = [seq_assay.upper() for seq_assay in all_seq_assays_raw]
+            all_seq_assays = assay_info_df.SEQ_ASSAY_ID.str.upper().unique()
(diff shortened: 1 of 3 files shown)


genie_registry/assay.py
```diff
@@ -141,7 +141,8 @@ def _validate(self, assay_info_df, project_id):
         warning = ""
 
         if process_functions.checkColExist(assay_info_df, "SEQ_ASSAY_ID"):
-            all_seq_assays = assay_info_df.SEQ_ASSAY_ID.unique()
+            all_seq_assays_raw = assay_info_df.SEQ_ASSAY_ID.unique()
```
thomasyu888: Try .upper().unique()

thomasyu888:
Need to add to tests

thomasyu888:
In the future, please lint in a separate commit or PR, because it's hard to find the actual code changes.  Also I didn't run black on the tests due to the awful formatting of the context managers but it doesn't actually look too bad now.

That being said there are failing tests though

hhunterzinck:
@thomasyu888 Tests were running fine before so linting did something.  Let me roll everything back and start from the beginning without linting

PR #437, opened 2022-02-03
fixes to seq-assay-id validation checks

### LINKED ISSUE #435: Make validation of SEQ_ASSAY_IDs case-insensitive

hhunterzinck:
Fixes #435

### DIFF
--- a/genie/process_functions.py
+++ b/genie/process_functions.py
@@ -876,18 +876,21 @@ def check_col_and_values(
                 final.extend(value.split(sep))
             check_values = pd.Series(final)
         if not check_values.isin(possible_values).all():
-            error = "{filename}: Please double check your {col} column.  " "This column must only be these values: {possible_vals}\n".format(
-                filename=filename,
-                col=col,
-                possible_vals=", ".join(
-                    [
-                        # This is done because of pandas typing.
-                        # An integer column with one NA/blank value
-                        # will be cast as a double.
-                        str(value).replace(".0", "")
-                        for value in possible_values
-                    ]
-                ),
+            error = (
+                "{filename}: Please double check your {col} column.  "
+                "This column must only be these values: {possible_vals}\n".format(
+                    filename=filename,
+                    col=col,
+                    possible_vals=", ".join(
+                        [
+                            # This is done because of pandas typing.
+                            # An integer column with one NA/blank value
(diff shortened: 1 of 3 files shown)


hhunterzinck:
All tests, including newly added test, pass but linting fails on two files:
genie_registry/assay.py
genie/process_functions.py

hhunterzinck:
All tests, including newly added test, pass but linting fails on one file:
genie/process_functions.py

hhunterzinck:
@thomasyu888  all tests and linting now pass

thomasyu888:
The processing needs to change for this too right? Or is there already functions to change to uppercase and replace the _ with -

hhunterzinck:
@thomasyu888 accounting for '_' to '-' now.  Any other formatting changes made to the SEQ_ASSAY_ID before upload to the Sample Clinical Database that we should account for?

thomasyu888:
Not sure. We just need to make sure the processing function in assay info does those transformations.

hhunterzinck:
@thomasyu888 Sure, we can make sure the processing functions for the assay files conduct the same functionality.  Would merging these changes to the validation function first affect the rest of the pipeline, though?  I just want to make sure we are not erroneously flagging these files for changes when we don't actually want the uploaders to make any changes.

thomasyu888:
The changes that are pulled into develop arent used in production yet. Will have to merge things into main and do a release for the changes to work.

That being said, there shouldn't be any issue except if these transformations aren't done in processing, then they will be uploaded into the assay table with underscores and lower case lettering.

hhunterzinck:
I think these changes are made already during processing, which is why the validation between the uploaded assay_information.yaml and the internal database was flagging the SEQ_ASSAY_IDs as issues.  See: https://github.com/Sage-Bionetworks/Genie/blob/0485b7e1599c4894714a92d5e12e99cc2b0cd02d/genie_registry/assay.py#L61

thomasyu888 (approved):

PR #441, opened 2022-02-16
Add sample class filter

### LINKED ISSUE #410: Add SAMPLE_CLASS filter to remove cfDNA samples for public release
https://github.com/Sage-Bionetworks/Genie/blob/master/genie/consortium_to_public.py#L128

thomasyu888:
Fixes #410

### DIFF
--- a/genie/database_to_staging.py
+++ b/genie/database_to_staging.py
@@ -999,8 +999,7 @@ def run_genie_filters(
     remove_seqdate_samples = seq_date_filter(
         clinicaldf, processing_date, consortium_release_cutoff
     )
-    logger.info("SAMPLE CLASS FILTER")
-    remove_sc_samples = sample_class_filter(clinical_df=clinicaldf)
+
     # Only certain samples are removed for the files that go into
     # staging center folder
     remove_center_consortium_samples = set(remove_mutationincis_samples).union(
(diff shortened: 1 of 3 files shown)


genie/database_to_staging.py
```diff
@@ -498,6 +498,19 @@ def seq_date_filter(clinicalDf, processingDate, consortiumReleaseCutOff):
...
+def sample_class_filter(clinical_df: pd.DataFrame) -> list:
+    """Filter samples by SAMPLE_CLASS
+
+    Args:
+        clinical_df (pd.DataFrame): Clinical dataframe
+
+    Returns:
+        list: List of samples to filter out
+    """
+    remove_samples = clinical_df["SAMPLE_ID"][clinical_df["SAMPLE_CLASS"] == "cfDNA"]
```
hhunterzinck: Do we know that "cfDNA" is the SAMPLE_CLASS value?

thomasyu888: There is only "Tumor" and "cfDNA" as allowed values here, so I'm assuming we don't want to filter out the 'Tumor' values?

hhunterzinck (approved):

hhunterzinck (approved):

PR #450, opened 2022-03-04
year or int of death is not applicable for living patients

### LINKED ISSUE #449: Modify check for DEAD variable in clinical file
if DEAD is False, then YEAR_DEATH and INT_DOD must be ‘Not Applicable’
https://www.synapse.org/#!Synapse:syn3380222/wiki/412290

hhunterzinck:
Fixes #449

### DIFF
--- a/genie_registry/clinical.py
+++ b/genie_registry/clinical.py
@@ -95,10 +95,10 @@ def _check_int_dead_consistency(clinicaldf: DataFrame) -> str:
     # Check that all string values are equal each other
     is_equal = all(clinicaldf.loc[is_str, "DEAD"] == clinicaldf.loc[is_str, "INT_DOD"])
     # If dead, int column can't be Not Applicable
-    # If alive, int column can't have values
+    # If alive, int column must be Not Applicable
     if (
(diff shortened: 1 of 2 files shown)


genie_registry/clinical.py
```diff
@@ -95,10 +95,10 @@ def _check_int_dead_consistency(clinicaldf: DataFrame) -> str:
     # Check that all string values are equal each other
     is_equal = all(clinicaldf.loc[is_str, "DEAD"] == clinicaldf.loc[is_str, "INT_DOD"])
     # If dead, int column can't be Not Applicable
-    # If alive, int column can't have values
+    # If alive, int column must be Not Applicable
```
thomasyu888, line 98: Just as a note here, I think we only need to check the interval and not the year, because we check for consistent strings between the INT and YEAR columns.  So by checking if INT = Not Applicable, the other check ensures that the YEAR = Not Applicable as well.  

Does that sound right to you?

thomasyu888 (approved):

genie_registry/clinical.py
hhunterzinck: yes, aligns with my thinking as well

PR #454, opened 2022-03-09
[IBCDPE-178] Create ValidationResult class 

thomasyu888:
prepare for detailed sample based errors to be returned


genie/example_filetype_format.py
```diff
@@ -6,6 +6,46 @@
...
+        Args:
+            errors (str): errors for the file
+            warnings (str): warning for the file
+            detailed_errors (pd.DataFrame, optional): Dataframe of detailed row based
+                                                      error messages. Defaults to None.
+        """
+        self.errors = errors
+        self.warnings = warnings
+        self.detailed = detailed
+        self.is_valid = errors == ""
```
hhunterzinck: Would it be safer to have `is_valid` as a function that performs an operation on `self.errors` to determine the validity?  In case the `errors` is updated in the object

thomasyu888: Yes, that is a really good suggestion. Will make changes

hhunterzinck (approved):

PR #460, opened 2022-03-23
[ETL-159] Use GENIE config dict for processing

thomasyu888:


hhunterzinck:
@thomasyu888 do you want me to review this one now or wait until the `create_config` branch is finalized?

thomasyu888:
They're both ready for review - minus the linting issue. Might be easier to review the create config branch first

genie/process_functions.py
```diff
@@ -1268,4 +1254,9 @@ def get_genie_config(
     # Add center configurations including input/staging synapse ids
     genie_config["center_config"] = center_mapping_df.to_dict("index")
 
+    genie_config['ethnicity_mapping'] = "syn7434242"
+    genie_config['race_mapping'] = "syn7434236"
+    genie_config['sex_mapping'] = "syn7434222"
+    genie_config['sampletype_mapping'] = "syn7434273"
```
hhunterzinck: Why not define these in project's `dbMapping` table directly?  I see they are specified as keys in the conftest.py: https://github.com/Sage-Bionetworks/Genie/blob/f135c98c5db7d1b9cebcc37b821578599072b9c6/tests/conftest.py#L62

thomasyu888: This is mainly because there are configurations for the test and staging projects and GENIE and I would need to add these mappings to those project's configuration tables.

hhunterzinck (approved):

PR #472, opened 2022-05-24
[IBCDPE-220] fix lgtm.com code scanning warnings/recommendations

thomasyu888:
LGTM.com is a website that does application security by doing code scanning with codeql and other tools to do code analysis. It produces "high" security vulnerabilities along with warnings and recommendations. It also has a feature that automatically reviews PRs for these concerns so that the code contributed doesn't contain issues.

* Specify empty dataframe with column headers to improve codebase
* Remove unused imports
* Do not set variables that aren't used
* Ignore some LGTM.com recommendations


genie_registry/bed.py
```diff
@@ -106,22 +106,23 @@ def _add_feature_type_tobeddf(filepath, featuretype):
...
-            "includeInPanel",
-            "clinicalReported",
-            "ID",
-            "SEQ_ASSAY_ID",
-        ]
+        beddf.columns = bed_columns
         beddf["Feature_Type"] = featuretype
     else:
-        beddf = pd.DataFrame()
+        beddf = pd.DataFrame(columns=bed_columns)
```
thomasyu888, line 126: Need to add in `Feature_Type` here

hhunterzinck (approved):

PR #505, opened 2023-01-24
[GEN-497] Remove fusion file from consortium release

thomasyu888:
**Purpose:** Remove fusion files so they are not released as part of the consortium release


genie/database_to_staging.py
```diff
@@ -1924,8 +1924,9 @@ def create_link_version(
             release_file["name"] != "data_clinical.txt" or release_type == "consortium"
         )
         is_gene_panel = release_file["name"].startswith("data_gene_panel")
+        is_depreciated_file = release_file["name"] in ["data_fusions.txt"]
```
thomasyu888, line 1835: This is great approach!  If we need a systematic way of handling deprecated files in the future, we can build off of this.

genie/database_to_staging.py
```diff
@@ -1746,16 +1746,16 @@ def stagingToCbio(
...
-        syn,
-        consortiumReleaseSynId,
-        genieVersion,
-        fusionSynId,
-        keepForCenterConsortiumSamples,
-        keepForMergedConsortiumSamples,
-        current_release_staging,
-        CENTER_MAPPING_DF,
-    )
+    # store_fusion_files(
```
thomasyu888: Can you delete this particular function and these commented code chunks to see how the code runs?

rxu17 (approved):
LGTM!

PR #506, opened 2023-02-04
[GEN-506] Support new synapser client - Update installation

thomasyu888:
* Use renv to install R packages in dockerfile
* Use new synapser client
* linted with black


Dockerfile
```diff
@@ -41,27 +41,43 @@ RUN apt-get update && apt-get install -y --allow-unauthenticated --no-install-re
...
+COPY renv/ renv/
+COPY renv.lock renv.lock
+RUN Rscript R/install_packages.R
 
 ENV CRYPTOGRAPHY_DONT_BUILD_RUST=true
-RUN Rscript R/install_packages.R
+COPY . .
+RUN echo "source('renv/activate.R')" >> .Rprofile
 
-RUN python3 -m pip install --no-cache-dir cython
```
rxu17, line 60: Do we not need cython anymore in the docker image?

rxu17 (approved):
Lgtm!

Dockerfile
thomasyu888: No, Im not sure why I added this previously, but now it's not needed because it gets installed as a dependency

PR #507, opened 2023-02-23
[GEN-496] remove fusions file support

rxu17:
**Purpose:** Removes the support for fusions file in main GENIE processing. This will result in uploaders receiving an invalid status for any fusions file they try to validate.


thomasyu888 (approved):
🛩️ LGTM!

PR #509, opened 2023-03-17
[GEN-428] Validate chromosome column in SV

rxu17:
**Purpose:** This PR adds the use of the chromosome validation function on the structural variant file format. 
This PR also adds the following tests:

- Adds tests in `test_validate.py` for the `_validate_chromosome` function and `ACCEPTED_CHROMOSOMES` variable
- Updates test in `test_sv.py` to have sample chromosome column and values to validate


tests/conftest.py
```diff
@@ -74,3 +74,8 @@ def genie_config():
 @pytest.fixture(scope="session")
 def syn():
     return mock.create_autospec(synapseclient.Synapse)
+
+
+@pytest.fixture(scope="session")
+def accepted_chromosomes():
```
thomasyu888: Nit: Do we need this if it is a global variable in `genie/validate.py`

thomasyu888 (approved):
🚀 going to approve, but minor comment! Good work!

tests/test_validate.py
```diff
@@ -205,6 +205,17 @@ def test_wrongfiletype_validate_single_file(syn):
         mock_determine_filetype.assert_called_once_with()
 
 
+def test_accepted_chromosomes_value():
```
thomasyu888, line 208: Great test!

tests/conftest.py
rxu17: I wasn't sure if it was good practice to use global variable directly in the tests. But I'll add a test to check for its value

tests/test_sv.py
thomasyu888, line 1: @rxu17 Can you add a test here for invalid chromosome to make sure that `_validate_chromosome` function is called?

Another solution is to `mock.patch` the function and use `assert_called_once_with` since you're already testing that individual function.

tests/test_sv.py
```diff
@@ -117,8 +120,28 @@ def test_validation_no_errors(self):
...
+    def test_validation__validate_chromosome_is_called(self):
+        """Tests that _validate_chromosome is called twice
+        for the two chromosome columns that sv files may have"""
+        sv_df = pd.DataFrame(
+            {
+                "sample_id": ["GENIE-SAGE-ID1-1", "GENIE-SAGE-ID2-1"],
+                "SITE1_CHROMOSOME": [1, 22],
+            }
+        )
+        with mock.patch(
```
thomasyu888: Nit: I know this isn't documented, but for consistency across the package, this is how mocking is done. (I also find it a little easier). Perhaps testing style should be documented in CONTRIBUTING.md

```
from unittest.mock import patch
patch.object(validate, "_validate_chromosome". return_value=("" ,"")
```

tests/test_sv.py
```diff
@@ -117,8 +120,28 @@ def test_validation_no_errors(self):
...
+            "genie.validate._validate_chromosome", return_value=("", "")
+        ) as validation__validate_chromosome_mock:
+            self.sv_cls._validate(sv_df)
+            assert validation__validate_chromosome_mock.call_count == 2, (
```
thomasyu888, line 144: Nit: This is fine for this purpose since we have tested this function.  BUT, for future reference, here's how I would tackle this to save time.

* Reduce the dataframe to length of one so the function is called only once.
* Do something like this to capture the[ actual calls](https://github.com/Sage-Bionetworks/Genie/blob/f6cea5b0794506549ff6cf7ce74290628ffc037b/tests/test_database_to_staging.py#L49-L58). (This is potentially over testing)

thomasyu888 (commented):
🔥 Just a few more comments.

PR #510, opened 2023-03-17
[GEN-429] Validate chromosome columns in bed file

rxu17:
Purpose: This PR adds the use of the chromosome validation function on the bed file format.
This PR also adds the following tests:

- Updates test in `test_sv.py` to have sample chromosome column and values to validate

depends on #509


thomasyu888 (approved):
🚀 LGTM!

PR #516, opened 2023-04-28
[GEN-636] Allow NAs/blanks for unrequired columns in SV

rxu17:
**Purpose:** This PR allows SV non-required columns to contain NAs/blanks. The ones of focus here are these:

- SITE1_CHROMOSOME
- SITE2_CHROMOSOME
- NCBI_BUILD

Additional transform functions were added to handle pandas' specialties of converting integer columns with NAs to floats, and converting NAs to strings when converting an entire column to the data type `str` so that the `_validate_chromosome` function could do the expected validation

Note that the `transform._convert_float_col_with_nas_to_int` function does assume any chromosome column was original integer data type and got converted to float data type when they have NAs in their values. This would mean that if any columns' original data type was float (not integer), they would pass validation as well.


thomasyu888 (approved):
🔥 excellent! We can put in a ticket to rename 'na_allowed' to 'allow_na'. I like that a lot better!

PR #517, opened 2023-05-06
Add in pre-commit

thomasyu888:


.pre-commit-config.yaml
```diff
@@ -0,0 +1,63 @@
...
+    language_version: python3
+
+- repo: https://github.com/PyCQA/bandit
+  rev: 1.7.5
+  hooks:
+  - id: bandit
+    args: ["-c", "pyproject.toml"]
+    additional_dependencies: ["bandit[toml]"]
+
+# - repo: https://github.com/asottile/blacken-docs
```
rxu17, line 39: Any reason we're not using these commented out checks/features?

rxu17 (approved):
LGTM!

.pre-commit-config.yaml
thomasyu888: Thats a good question, I think there are errors that pop up with the commented out pre-commit tags.  Especially `mypy` would be a huge issue...

PR #518, opened 2023-05-06
[GEN-81] Deprecate support for passwords, and authentication parameters in cli

thomasyu888:
Deprecate support for
* passwords
* pemfile
* cli parameters for authentication

Will need to provide documentation for users to set up their synapse client via env variable or ~/.synapseConfig using personal access token.


rxu17 (approved):
LGTM! The amount of code removed due to the synapse login implementation :o

PR #519, opened 2023-05-17
[GEN-694] Remove fusions from dashboarding

thomasyu888:
* Remove fusions from dashboarding
* Remove models

### DIFF
--- a/genie/dashboard_table_updater.py
+++ b/genie/dashboard_table_updater.py
@@ -129,13 +129,13 @@ def update_cumulative_sample_table(
     clinical_counts["Total"] = sum(clinical_counts)
     clinical_counts.name = "Clinical"
 
-    fusion_ent = syn.get(file_mapping["fusion"], followLink=True)
-    fusiondf = pd.read_csv(fusion_ent.path, sep="\t", comment="#")
-    fusiondf.columns = [i.upper() for i in fusiondf.columns]
-    fusion_counts = fusiondf["CENTER"][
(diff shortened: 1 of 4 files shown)


genie/dashboard_table_updater.py
```diff
@@ -151,7 +151,7 @@ def update_cumulative_sample_table(
     seg_counts["Total"] = sum(seg_counts)
 
     total_counts = pd.DataFrame(clinical_counts)
-    total_counts["Fusions"] = fusion_counts
+    total_counts["Fusions"] = sv_counts
```
rxu17, line 152: Any reason we don't update the `total_counts["Fusions"]` to `total_counts["Structural_Variants"]` or something like that? Similar question for Lines 266, 459, 465?

rxu17 (approved):
LGTM (after the linting resolution) ! Just a question

genie/dashboard_table_updater.py
thomasyu888: @rxu17 That is a great question.  We should create a ticket to rename "Fusions" to "Structural_Variants", BUT... All fusions are structural variations and vice versa.

rxu17: Makes sense!
I just wasn't sure if it mattered that it affected the appearance of the `cumulative_sample_table` in synapse in that it still used `Fusions` as the naming convention

thomasyu888: Ya... the tables themselves still use `Fusions` as the column header so thats why. We would need to update the headers in the Synapse tables along with the code which is why those are the same.

PR #520, opened 2023-05-17
[GEN-560] Add ancillary files for cross-validation

rxu17:
**Purpose:** This passes ancillary files as an attribute to the `ValidationHelper` and `FileTypeFormat` classes so that we can implement cross validation for specific file formats in the future


genie/example_filetype_format.py
```diff
@@ -60,10 +60,20 @@ class FileTypeFormat(metaclass=ABCMeta):
 
     _validation_kwargs = []
 
-    def __init__(self, syn, center, genie_config=None):
+    def __init__(self, syn, center, genie_config=None, ancillary_files=None):
```
thomasyu888: Nit: Add typing hinting

genie/validate.py
```diff
@@ -108,7 +111,11 @@ def validate_single_file(self, **kwargs):
                 syn=self._synapse_client,
                 center=self.center,
                 genie_config=self.genie_config,
+                ancillary_files=self.ancillary_files,
             )
+            print("------Printing ancillary files------")
+            print(validator.ancillary_files)
```
thomasyu888: Nit: I think the PR to actually use the ancillary files actually covers this print statement, originally, I was actually trying to ensure that the ancillary files are passed to the classes that extend `example_filetype_format` but I think we are ok

thomasyu888 (approved):
🔥 LGTM! Sorry - lagged on the review, just some comments, but pre-approving

PR #521, opened 2023-05-17
Add mypy pre-commit

thomasyu888:
* Fix existing mypy issues by fixing the output of `pre-commit run --all-files`
* Allow for new code to run mypy as long as type hints are leveraged

depends on #524 (This is so that we don't merge this)


rxu17 (approved):
LGTM! 🎨

PR #522, opened 2023-05-18
[GEN-562] Cross-validate maf with clinical sample files

rxu17:
**Purpose:** This PR adds further cross-validation functionality by adding the ability to cross-validate ids in the maf files with the clinical sample files. 

This uses the newly introduced `ancillary_files` attribute in the specific filetypeformat classes to cross validate.

### DIFF
--- a/genie/example_filetype_format.py
+++ b/genie/example_filetype_format.py
@@ -276,7 +276,7 @@ class FileTypeFormat(metaclass=ABCMeta):
         """
         errors = ""
         warnings = ""
-        logger.info("NO VALIDATION for %s files" % self._fileType)
+        logger.info("NO CROSS-VALIDATION for %s files" % self._fileType)
         return errors, warnings
 
     def validate(self, filePathList, **kwargs) -> ValidationResults:
@@ -310,13 +310,15 @@ class FileTypeFormat(metaclass=ABCMeta):
 
         if not errors:
             logger.info("VALIDATING %s" % os.path.basename(",".join(filePathList)))
-            errors_validate, warnings_validate = self._validate(df, **mykwargs)
+            errors, warnings = self._validate(df, **mykwargs)
             logger.info(
                 "CROSS-VALIDATING %s" % os.path.basename(",".join(filePathList))
             )
-            errors_cross_validate, warnings_cross_validate = self._cross_validate(df)
-            errors = f"{errors_validate}\n{errors_cross_validate}"
-            warnings = f"{warnings_validate}\n{warnings_cross_validate}"
+            # only cross-validate if validation passes
+            if not errors:
+                errors_cross_validate, warnings_cross_validate = self._cross_validate(df)
+                errors = f"{errors}\n{errors_cross_validate}"
+                warnings = f"{warnings}\n{warnings_cross_validate}"
             
 
         result_cls = ValidationResults(errors=errors, warnings=warnings)
--- a/genie/example_filetype_format.py
+++ b/genie/example_filetype_format.py
@@ -265,8 +265,8 @@ class FileTypeFormat(metaclass=ABCMeta):
                 errors_cross_validate, warnings_cross_validate = self._cross_validate(
                     df
                 )
-                errors = f"{errors_cross_validate}\n"
-                warnings = f"{warnings}\n{warnings_cross_validate}\n"
+                errors += errors_cross_validate
+                warnings += warnings_cross_validate
 
         result_cls = ValidationResults(errors=errors, warnings=warnings)
         return result_cls
--- a/genie/example_filetype_format.py
+++ b/genie/example_filetype_format.py
@@ -200,6 +200,77 @@ class FileTypeFormat(metaclass=ABCMeta):
         logger.info("NO VALIDATION for %s files" % self._fileType)
         return errors, warnings
 
+    def cross_validate_ids_between_two_files(
+        self,
+        df1: pd.DataFrame,
+        df2_file_name: str,
+        id_to_check: str,
+    ) -> tuple:
+        """Check that all the identifier(s) (ids) in one
+        file exists in the other file
+
+        Args:
+            df1 (pd.DataFrame): file to use as base of check
+            df2_file_name (str): file name of the file to cross-validate against
+            id_to_check (str): name of column to check values for
+
+        Returns:
+            tuple: The errors and warnings as a file from cross-validation.
+                   Defaults to blank strings
+        """
+        errors = ""
+        warnings = ""
+        if df2_file_name in self.ancillary_files:
+            df2_filepath = self.ancillary_files[df2_file_name]["entity"].path
+            df2_filetype = self.ancillary_files[df2_file_name][
+                "filetypeformat_object"
+            ]._fileType
+
+            try:
+                df2 = self.ancillary_files[df2_file_name][
+                    "filetypeformat_object"
+                ].read_file(df2_filepath)
+            except Exception as e:
+                errors = f"The file(s) to be cross-validated against ({df2_filepath}) cannot be read. Original error: {str(e)}"
+                warnings = ""
+
+            # standardize case
+            df2.columns = [col.upper() for col in df2.columns]
+
+            if id_to_check in df2.columns and id_to_check in df1.columns:
+                # check to see if the ids are equal
+                if set(df1[id_to_check]) != set(df2[id_to_check]):
+                    errors = (
+                        f"The {id_to_check}s between {self._fileType} and "
+                        f"{df2_filetype} are not equal."
+                    )
+                    warnings = ""
+            else:
+                errors = ""
+                warnings = f"{id_to_check} doesn't exist in {self._fileType} or {df2_filetype}. No cross-validation will be done."
+        else:
+            errors = ""
+            warnings = f"{df2_file_name} doesn't exist. No cross-validation will be done between"
+            f"{self._fileType} and {df2_filetype}"
+        return errors, warnings
+
+    def _cross_validate(self, df: pd.DataFrame) -> tuple:
+        """
+        This is the base cross-validation function.
+        By default, no cross-validation occurs.
+
+        Args:
+            df (pd.DataFrame): A dataframe of the file
+
+        Returns:
+            tuple: The errors and warnings as a file from cross-validation.
+                   Defaults to blank strings
+        """
+        errors = ""
+        warnings = ""
+        logger.info("NO VALIDATION for %s files" % self._fileType)
+        return errors, warnings
+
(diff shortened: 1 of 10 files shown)


genie/example_filetype_format.py
```diff
@@ -200,6 +200,77 @@ def _validate(self, df: pd.DataFrame, **kwargs) -> tuple:
         logger.info("NO VALIDATION for %s files" % self._fileType)
         return errors, warnings
 
+    def cross_validate_ids_between_two_files(
```
thomasyu888: To reduce complexity in `example_filetype_format.py` a quick thought I had was to add this to `validate.py`

```
def check_values_between_two_df(df1, df2, col1, col2)
   if df1[col1].isin(df2[col2]):
      ....
```

thomasyu888:
This is a wonderful start, let's discuss more soon

genie_registry/maf.py
```diff
@@ -277,6 +277,25 @@ def _validate(self, mutationDF):
 
         return total_error.getvalue(), warning.getvalue()
 
+
+    def _cross_validate(self, mutationDF: pd.DataFrame) -> tuple:
```
thomasyu888, line 280: First step:
1. You have a list of ancillary files
2. Loop through them to find "data_clinical_supp_{center}.txt" OR "data_clinical_supp_sample_{center}.txt"
3. Once you find this ^, `from clinical import clinical, clinical()._read_data(...)`, <- if this doesn't work, hard code `try catch pd.read_csv(sep="\t")`
4. If _validate() has no errors, then you can assume maf passed validation, and only check for if the sample file has the right headers, etc, maf['TUMOR_SAMPLE_BARCODE']` exists.  
5. pass the maf df and sample df into the comparison function

genie/example_filetype_format.py
```diff
@@ -231,7 +302,14 @@ def validate(self, filePathList, **kwargs) -> ValidationResults:
 
         if not errors:
             logger.info("VALIDATING %s" % os.path.basename(",".join(filePathList)))
-            errors, warnings = self._validate(df, **mykwargs)
+            errors_validate, warnings_validate = self._validate(df, **mykwargs)
+            logger.info(
+                "CROSS-VALIDATING %s" % os.path.basename(",".join(filePathList))
+            )
+            errors_cross_validate, warnings_cross_validate = self._cross_validate(df)
```
thomasyu888: If errors_validate: don't do cross validate?

genie/input_to_database.py
```diff
@@ -165,6 +166,75 @@ def _get_status_and_error_list(valid, message, entities):
...
+            validator = validate.GenieValidationHelper(
+                syn=syn,
+                project_id=project_id,
+                center=center,
+                entitylist=[ent.path],
+                format_registry=format_registry,
+                genie_config=genie_config,
+                ancillary_files=None,
+            )
+            filetype = validator.file_type
```
thomasyu888, line 227: Two paths forward, either handle the custom clinical class, OR... comment this section out and just return a dictionary of entities:
```
[{"filename": {"entity": entity}}]
```

genie/input_to_database.py
```diff
@@ -165,6 +166,109 @@ def _get_status_and_error_list(valid, message, entities):
     return input_status_list, invalid_errors_list
 
 
+def get_ancillary_files(
```
thomasyu888, line 169: @rxu17 I didn't realize this until right now, another option is to pass in the list of lists from the `get_center_files` to use in this function so we don't crawl through the files again

There is quite a bit of added complexity here that I'm just not sure the code is ready to handle due to not having `sample` and `patient` classes.

I would opt for simplicity over complexity - since the only benefit right now only seems to be the ability to read data-frame, it might be easier to tackle this benefit after we can do it correctly.

rxu17: In the end, decided to comment out this function and instead use the `extract.get_center_files`.

rxu17:
@thomasyu888 I updated the structure so this is what the cross validation will generally look like for other cross validation tickets we use. Let me know what you think (I can also move `cross_validate_ids_between_two_files` to the `validate` class), and I'll move forward with this.

- Copy the `get_dataframe` from the individual filetypeformat class to `process_functions` and rename it as specific to the filetypeformat class it belongs to
- Each filetypeformat class's `_cross_validate` function will loop through `ancillary_files` (which is just the result of `extract.get_center_files` now) and get the file that we are trying to cross validate against so we can read it in
- Do specific cross-validation specific to the filetypeformat object pair

When we do change the structure of `get_center_files`/separate out the clinical files, we should be able to just modify each filetypeformat class's `_cross_validate` function and remove the places that loop through the list of lists as well as just removing the `get_dataframe` function we copied to `process_functions` and replace it with the class specific one. Everything else should stay the same.

I might just make the section of the code where we loop through the `ancillary_files` as its own function so easier to remove in the future.

genie/example_filetype_format.py
```diff
@@ -208,6 +208,64 @@ def _validate(self, df: pd.DataFrame, **kwargs) -> tuple:
         logger.info("NO VALIDATION for %s files" % self._fileType)
         return errors, warnings
 
+    def cross_validate_ids_between_two_files(
```
thomasyu888: I still prefer this to be moved into `validate.py` even though it might be used many times.  Mainly because this is a `@staticmethod` and doesn't use `self` at all

genie_registry/maf.py
```diff
@@ -277,6 +277,58 @@ def _validate(self, mutationDF):
...
+                        df2=clinical_sample_df,
+                        df2_filename=clinical_file_names,
+                        id_to_check=id_to_check,
+                    )
+                else:
+                    errors = ""
+                    warnings = f"{id_to_check} doesn't exist in the maf or clinical file(s). No cross-validation will be done."
+        else:
+            errors = ""
+            warnings = f"The clinical file(s) doesn't exist. No cross-validation will be done."
```
thomasyu888: This can be an error.

genie_registry/maf.py
```diff
@@ -277,6 +277,58 @@ def _validate(self, mutationDF):
...
+                )
+                warnings = ""
+
+            if not errors:
+                if process_functions.checkColExist(
+                    mutationDF, id_to_check
+                ) and process_functions.checkColExist(clinical_sample_df, id_to_check):
+                    errors, warnings = self.cross_validate_ids_between_two_files(
+                        df1=mutationDF,
+                        df1_filename=maf_filename,
```
thomasyu888: The maf file has the sample column at `Tumor_Sample_Barcode` which is different from `SAMPLE_ID`

thomasyu888 (changes requested):
Nice!  This is starting to shape together!

genie/process_functions.py
```diff
@@ -22,6 +22,42 @@
 SCRIPT_DIR = os.path.dirname(os.path.abspath(__file__))
 
 
+def get_clinical_dataframe(filePathList):
```
thomasyu888: Nit: Can we add light docstrings. This will eventually be refactored anyhow.

genie_registry/maf.py
```diff
@@ -277,6 +277,63 @@ def _validate(self, mutationDF):
...
+                clinical_sample_df = process_functions.get_clinical_dataframe(
+                    filePathList=clinical_file_paths
+                )
+            except Exception as e:
+                errors = f"The clinical file(s) cannot be read. No cross-validation will be done. Original error: {str(e)}"
+                warnings = ""
+
+            if not errors:
+                if process_functions.checkColExist(
+                    mutationDF, "SAMPLE_ID"
```
thomasyu888: You flipped the sample columns in maf and sample. Mutation df should have tumor sample barcode and clinical sample should have sample id

genie_registry/maf.py
```diff
@@ -277,6 +277,63 @@ def _validate(self, mutationDF):
...
+                        df1_filename=maf_filename,
+                        df1_id_to_check="SAMPLE_ID",
+                        df2=clinical_sample_df,
+                        df2_filename=clinical_file_names,
+                        df2_id_to_check="TUMOR_SAMPLE_BARCODE",
+                    )
+                else:
+                    errors = ""
+                    warnings = (
+                        "SAMPLE_ID doesn't exist in the maf or TUMOR_SAMPLE_BARCODE doesn't exist clinical file(s)."
```
thomasyu888: You don't need to do the sample Id check in the maf because that is done in the row based validation step.

If initial variation fails, no cross file validation is done.

thomasyu888 (commented):
Great work - just some minor things.

tests/test_example_filetype_format.py
```diff
@@ -0,0 +1,74 @@
+from unittest.mock import patch
```
thomasyu888, line 1: Great tests!

genie_registry/maf.py
```diff
@@ -277,6 +277,59 @@ def _validate(self, mutationDF):
 
         return total_error.getvalue(), warning.getvalue()
 
+    def _cross_validate(self, mutationDF: pd.DataFrame) -> tuple:
```
thomasyu888, line 280: Nit: let's try to do snakecase with everything moving forward.

thomasyu888 (approved):
🔥 awesome work! I'm going to pre-approve, but can you paste different samples of the error email for @Chelsea-Na to provide any of her suggestions?

genie/example_filetype_format.py
```diff
@@ -240,6 +257,16 @@ def validate(self, filePathList, **kwargs) -> ValidationResults:
         if not errors:
             logger.info("VALIDATING %s" % os.path.basename(",".join(filePathList)))
             errors, warnings = self._validate(df, **mykwargs)
+            logger.info(
```
thomasyu888: Nit: Should this be after the "if not errors"?

genie/validate.py
```diff
@@ -251,3 +251,67 @@ def _perform_validate(syn, args):
...
+
+    # check to see if df1 ids are present in df2
+    if not set(df1[df1_id_to_check]) <= set(df2[df2_id_to_check]):
+        errors = (
+            f"At least one {df1_id_to_check} in your {df1_filename} file "
+            f"does not exist as a {df2_id_to_check} in your {df2_filename} file. "
+            "Please update your file(s) to be consistent.\n"
+        )
+
+    return errors, warnings
```
thomasyu888, line 317: Nit: In the future, it  might be nice to use `namedtuples` so we aren't relying on the ordering of the tuple.

thomasyu888 (approved):
🔥 Nice!  I'm excited about this!

thomasyu888 (approved):
🔥 Nice!  I'm excited about this!

PR #523, opened 2023-05-30
[GEN-561] Add cross validation to check for seq-assay-id bed file pair

rxu17:
**Purpose:** This PR adds a cross validation function to check that there is a bed file for every `SEQ-ASSAY-ID` value in the clinical file.

This PR's merge is blocked by #522 (The branch will update to `develop` once #522 is merged into `develop`)

Depends on #522

### DIFF
--- a/genie/example_filetype_format.py
+++ b/genie/example_filetype_format.py
@@ -276,7 +276,7 @@ class FileTypeFormat(metaclass=ABCMeta):
         """
         errors = ""
         warnings = ""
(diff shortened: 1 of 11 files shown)


thomasyu888 (approved):
🔥 fantastic! Going to let you do the merges. 

Note: you can use "depends on" like I added for you to add a auto check. Dpulls is a plugin used for our repos at sage.  Notice how the checks fail.

PR #524, opened 2023-06-01
[GEN-706] Cross-validate clinical assay information

rxu17:
**Purpose:** This PR adds a cross validation function to check that there is a matching `SEQ_ASSAY_ID` value in the `assay_information.yaml` file for every `SEQ_ASSAY_ID` value in the clinical file.

Depends on #523


thomasyu888 (approved):
🦖 nice! Going to pre-approve but let's wait for Chelsea's input on error messages!

PR #526, opened 2023-06-03
Do not cross-file validate if ancillary files not present

thomasyu888:
Currently there is a bug introduced when users validate via the cli because ancillary files will be None by default. 

This will ensure that cross validation doesn't happen if there are no ancillary files.  Support for ancillary files via the cli can come later.

### DIFF
--- a/genie/example_filetype_format.py
+++ b/genie/example_filetype_format.py
@@ -2,9 +2,10 @@
 This contains the GENIE model objects
 """
 from abc import ABCMeta
+from dataclasses import dataclass
 import logging
 import os
-from typing import List
+from typing import List, Optional
 
 import pandas as pd
 
@@ -12,20 +13,13 @@ import pandas as pd
 logger = logging.getLogger(__name__)
 
 
+@dataclass
 class ValidationResults:
     """Validation results"""
 
-    def __init__(self, errors: str, warnings: str, detailed: str = None) -> None:
-        """
-        Args:
-            errors (str): errors for the file
-            warnings (str): warning for the file
-            detailed_errors (pd.DataFrame, optional): Dataframe of detailed row based
-                                                      error messages. Defaults to None.
-        """
-        self.errors = errors
-        self.warnings = warnings
-        self.detailed = detailed
+    errors: str
+    warnings: str
+    detailed: Optional[str] = None
 
     def is_valid(self) -> bool:
         """True if file is valid"""
--- a/genie/example_filetype_format.py
+++ b/genie/example_filetype_format.py
@@ -258,7 +258,7 @@ class FileTypeFormat(metaclass=ABCMeta):
             logger.info("VALIDATING %s" % os.path.basename(",".join(filePathList)))
             errors, warnings = self._validate(df, **mykwargs)
             # only cross-validate if validation passes
-            if not errors:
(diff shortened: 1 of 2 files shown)


genie/example_filetype_format.py
```diff
@@ -258,7 +258,7 @@ def validate(self, filePathList, **kwargs) -> ValidationResults:
             logger.info("VALIDATING %s" % os.path.basename(",".join(filePathList)))
             errors, warnings = self._validate(df, **mykwargs)
             # only cross-validate if validation passes
-            if not errors:
+            if not errors and self.ancillary_files is not None:
```
thomasyu888: This will also be an issue if ancillary files is an empty list.  Maybe... this should just be `and self.ancillary_files`

Chelsea-Na (approved):

rxu17 (approved):
LGTM! 🐛

genie/example_filetype_format.py
thomasyu888: @rxu17 does the thumbs up mean you think I should change this line?

rxu17: Sorry, I read your comment incorrectly. Yes, I think `not self.ancillary_files` would be good here then

rxu17: @thomasyu888
Sorry I read your comment incorrectly.
So `not self.ancillary_files` would handle the scenarios where `ancillary_files = None, []`

But we would still need `ancillary_files` to be a list of lists. 
Would something like this help with that:

```suggestion
           is_list_of_lists = all(isinstance(entities, list) for entities in ancillary_files)
           if not errors and isinstance(ancillary_files, list):
               if is_list_of_lists:
                    do stuff
```

tests/test_example_filetype_format.py
```diff
@@ -8,9 +8,38 @@
...
 def filetype_format_class(syn):
+    yield FileTypeFormat(syn, "SAGE", ancillary_files=[["mocked"]])
+
+
+@pytest.fixture
+def filetype_format_class_no_ancillary(syn):
     yield FileTypeFormat(syn, "SAGE")
 
 
+def test_that_validate_no_cross_file_without_ancillary(
```
rxu17: Would add more tests here if we update what we are checking for in the `ancillary_files`

rxu17 (commented):
Left some comments

genie/example_filetype_format.py
thomasyu888: Hmm, but would this pass?

```
ancillary_files=[[]]
```

rxu17: Ah, yeah sorry `parse_file_info_in_nested_list` expects it to not have any empty list in the `ancillary_files` so even [[]], [[],[],[]] or [[], ["something"], []], wouldn't work

Could use then `all(entities and isinstance(entities, list) for entities in ancillary_files)` for the `is_list_of_lists` check above?

thomasyu888: @rxu17 Thanks for all the feedback.  After thinking about this some more, I think it makes most sense to refactor to this to use pydantic validators, or have a method to validate the class attributes. 

I am adding mypy static type checking in another branch which will help us catch some of these errors as well, but I'm opting to not over complicate this and pull this in as it stands.

PR #530, opened 2023-07-07
[GEN-756] Fix maf validation

rxu17:
**Purpose:** This adds a helper function to `maf.py`'s `_get_dataframe` function to allow for handing of files where it gets read in as mixed dtypes to avoid doing so.

Note that we still need to run a larger maf dataset that has mixed dtypes prior to validating through validation as part of an integration test.

### DIFF
--- a/tests/test_transform.py
+++ b/tests/test_transform.py
@@ -1,4 +1,5 @@
 """Test genie.transform module"""
+import os
 from io import BytesIO
 from unittest.mock import patch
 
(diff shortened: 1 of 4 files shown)


tests/test_transform.py
```diff
@@ -73,3 +86,75 @@ def test_that__convert_float_col_with_nas_to_int_does_nothing_with_str_data(self
...
         assert result[0] == "Val1"
         assert pd.isna(result[1])
+
+
+def read_csv_side_effect(**kwargs):
+    if kwargs["low_memory"] == True:
+        raise pd.errors.DtypeWarning("some_warning")
+
+
+class TestConvertMixedDtypes:
```
thomasyu888, line 97: Just a discussion: but is this too tied to the implementation of the functions?  Does `_get_dataframe` cover these tests?

thomasyu888 (approved):
🔥 I'm going to pre-approve - thanks for doing this!  Just a discussion point.

tests/test_transform.py
rxu17: My reasoning behind creating it as a separate function was so we didn't have to mock out too many other things and could just focus on testing the implementation of the `_convert_df_with_mixed_dtypes`.
The tests test for:

- The scenario that a dataset will only need to be read using `low_memory = False`
- The scenario that a dataset will need to be read in twice, once using `low_memory = False` and the other using `low_memory = True`

But I think I see your point -- the test that is checking the output if it's directly equal to a specific dataframe is technically already covered by the `_get_dataframe` tests.

thomasyu888: I am ok with leaving as if.  If we ever end up changing the implementation of that function, and this test fails for some reason, as long as the `_get_dataframe` test passes, we might be able to remove it.

PR #532, opened 2023-07-10
[GEN-498] Remove fusion files from public release

rxu17:
**Purpose:** Remove fusion files from public release

### DIFF
--- a/genie/consortium_to_public.py
+++ b/genie/consortium_to_public.py
@@ -189,7 +189,7 @@ def consortiumToPublic(
     mapping = extract.get_syntabledf(syn=syn, query_string="SELECT * FROM syn9621600")
     genePanelEntities = []
     for entName, entId in consortiumRelease[2]:
-        is_depreciated_file = entName in ["data_fusions.txt"]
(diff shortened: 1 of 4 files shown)


genie/consortium_to_public.py
```diff
@@ -192,6 +189,7 @@ def consortiumToPublic(
     mapping = extract.get_syntabledf(syn=syn, query_string="SELECT * FROM syn9621600")
     genePanelEntities = []
     for entName, entId in consortiumRelease[2]:
+        is_depreciated_file = entName in ["data_fusions.txt"]
```
thomasyu888: Nit: 😆 depreciated? we are now de-valuing file formats. /s but also, its very close in spelling with "deprecated"

thomasyu888 (changes requested):
🔥 I am very excited about this update.  Can we also remove this: https://github.com/Sage-Bionetworks/Genie/blob/4a4df423f2a06480d85b54bf0726c8bbb96af42c/genie/create_case_lists.py#L297.

Thanks!

thomasyu888 (approved):
🦖 nice! Going to approve!

PR #533, opened 2023-07-11
[GEN-738] Add variant position validation for maf file

rxu17:
**Purpose:** This adds to `validate.py` a check for start and end positions for variant, and that the start position must be < end position. This is implemented for `maf` files.

### DIFF
--- a/genie/validate.py
+++ b/genie/validate.py
@@ -318,3 +318,34 @@ def check_values_between_two_df(
         )
 
     return errors, warnings
+
+
+def check_variant_start_and_end_positions(
+    input_df: pd.DataFrame, start_pos_col: str, end_pos_col: str, filename: str
+) -> tuple:
+    """Checks that a variant's start position is less than its
+        end position
+
+    Args:
+        input_df (pd.DataFrame): Input data to check positions
+        start_pos_col (str): Name of the start position column
+        end_pos_col (str): Name of the end position column
+        filename (str): Name of the file
+
+    Returns:
+        tuple: The errors and warnings from the position validation
+               Defaults to blank strings
+
+    """
+    errors = ""
+    warnings = ""
+
+    if any(input_df[start_pos_col] > input_df[end_pos_col]):
+        errors = (
+            f"{filename}: Your variants file has record(s) that have an end position value "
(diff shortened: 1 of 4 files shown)


genie/validate.py
```diff
@@ -318,3 +318,34 @@ def check_values_between_two_df(
...
+        tuple: The errors and warnings from the position validation
+               Defaults to blank strings
+
+    """
+    errors = ""
+    warnings = ""
+
+    if any(input_df[start_pos_col] > input_df[end_pos_col]):
+        errors = (
+            f"{filename}: Your variants file has record(s) that have an end position value "
```
thomasyu888: @Chelsea-Na what do you think of this error message?

> Your variants file has record(s) that have an end position value less than the start position value. When we annotate through the genome-nexus-annotation-pipeline, the records with this position discrepancy will show a blank reference and variant allele. Please update your file to be consistent.

genie_registry/maf.py
```diff
@@ -275,6 +275,18 @@ def _validate(self, mutationDF):
             )
             total_error.write(error)
 
+        # only check end position as start position is required col
+        if process_functions.checkColExist(
```
thomasyu888: I didn't noticed this, but perhaps we should add the `START_POSITION` and `END_POSITION` in the numerical value check above to make sure these are numerical values.  This way you don't need to update the `check_variant_start_and_end_positions` function to convert values to numerical. 

There will be a validation error if a non-numerical value gets passed into START or END position.

rxu17: Ah I was wondering about that. I saw that happening for another file format namely `bed` so I thought `maf` didn't need it. I can add that.

genie/validate.py
Chelsea-Na: Your variants file has record(s) that have an end position value less than the start position value. Please update your file to be consistent. When we annotate using the genome-nexus-annotation-pipeline, the records with this position discrepancy will show a blank reference and variant allele.

thomasyu888 (approved):
🔥 awesome - please update with what Chelsea suggested as a new error message but I'm going to pre-approve.

PR #535, opened 2023-09-14
[GEN-851] Hotfix pin cbioportal repo version

rxu17:
**Purpose:** Running into the issue of a missing package when pulling the latest repo version from cBioPortal whilest running the public release. We will pin the version of cbioportal repo to an earlier one of `v5.3.19` in our dockerfile.


thomasyu888 (approved):
🐛 Thanks for the fix!

PR #536, opened 2023-09-15
[GEN-846] & [GEN-845] Ignore case and allow underscores in cross-validate

rxu17:
**Purpose:** We would like to ignore case and allow underscores when cross-validating `SEQ_ASSAY_ID`s in clinical files against assay_information files' `SEQ_ASSAY_ID` values and bed file names.

This involves an addition of a new function to `validate.py` called `standardize_string_for_validation`. This function is called in all the affected clinical-assay_information and clinical-bed cross-validation functions within the `clinical.py` script.

Example) Something like SAGE_tEst-1 should compare as **equal** to something like Sage-test-1

This also completes JIRA ticket: [GEN-845](https://sagebionetworks.jira.com/browse/GEN-845)

[GEN-845]: https://sagebionetworks.jira.com/browse/GEN-845?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/genie/validate.py
+++ b/genie/validate.py
@@ -255,13 +255,18 @@ def _perform_validate(syn, args):
 
 
 def parse_file_info_in_nested_list(
-    nested_list: List[List[synapseclient.Entity]], search_str: str
+    nested_list: List[List[synapseclient.Entity]],
+    search_str: str,
+    ignore_case: bool = False,
+    allow_underscore: bool = False,
 ) -> dict:
     """Parses for a name and filepath in a nested list of Synapse entity objects
 
     Args:
         nested_list (list[list]): _description_
         search_str (str): the substring to look for in the files
+        ignore_case (bool, optional): whether to perform case-insensitive comparison.
+        allow_underscore (bool, optional): whether to treat underscores as equivalent to dashes.
 
     Returns:
         dict[dict]: files found,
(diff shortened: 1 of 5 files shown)


tests/test_validate.py
```diff
@@ -603,3 +722,31 @@ def test_that_check_variant_start_and_end_positions_returns_expected(
     )
     assert errors == expected_errors
     assert warnings == expected_warnings
+
+
+def test_that_standardize_string_for_validation_ignores_case():
```
thomasyu888: Nit:  I think these standardize validation tests can be written with parametrize.

tests/test_validate.py
```diff
@@ -507,34 +507,151 @@ def test_that_parse_file_info_in_nested_list_returns_expected(
...
+            "",
+            True,
+            False,
+        ),
+        (
+            pd.DataFrame({"ID": ["TEST_1", "TEST_2", "TEST-3"]}),
+            pd.DataFrame({"ID2": ["TEST_1", "TEST-2", "TEST_3"]}),
+            "",
+            "",
+            False,
```
thomasyu888, line 627: Did you think we need test cases for when we don't enable these that they return the error strings?

genie/validate.py
```diff
@@ -348,3 +387,28 @@ def check_variant_start_and_end_positions(
             "position discrepancy will show a blank reference and variant allele.\n"
         )
     return errors, warnings
+
+
+def standardize_string_for_validation(
```
thomasyu888, line 395: This is good foresight, we do a lot of string standardization in genie during processing. It might help in the long term to have a general string standardization strategy across all files, but this is fine for now.

thomasyu888 (commented):
Nice work!  Just some comments.

genie/validate.py
rxu17: Yup, agreed

tests/test_validate.py
rxu17: Might be mostly covered by the original test cases but probably wouldn't hurt to add another one for just strings

thomasyu888 (approved):
💯 Going to pre-approve!  This is great

PR #537, opened 2023-09-26
[GEN-834] Disable data guide generation

rxu17:
**Purpose:** Data guide generation is a separate process and the code to create it in the main genie process will need to be disabled. Ran test consortium and public releases.


thomasyu888 (approved):
👍🏼 LGTM! This will simplify the docker build once we get to removing the "dead code"

PR #538, opened 2023-10-10
[GEN-905] - Change error of start position > end position to a warning

thomasyu888:
Genome Nexus now handles this issue, so instead of an error, we just throw a warning


Chelsea-Na (approved):
Thank you! Looks good.

PR #539, opened 2023-11-03
[GEN-809] Validate allele columns

rxu17:
**Purpose:** There is no current allele validation so this PR adds it for the VCF and MAF file formats. Each format has its own list of accepted alleles. 

**Changes:** I added two general allele validation and allele validation message functions to `genie/validate.py`:
- [get_invalid_allele_rows](https://github.com/Sage-Bionetworks/Genie/blob/gen-809-validate-allele/genie/validate.py#L421)
- [get_allele_validation_message](https://github.com/Sage-Bionetworks/Genie/blob/gen-809-validate-allele/genie/validate.py#L466)

so that they can be more general and used by both vcf and maf file formats (and any relevant future file formats). I separated out the retrieval of the rows with the invalid allele values and the retrieval of the error message to make it easier for testing and also easier to implement row-based validation in the future if we would like (or keep our more high level validation).

**Special Considerations:** 
NA handling: Based on our discussion, we will wait to handle NAs (default solution right now is to flag NAs as invalid). Because I'm using regex to match on the allele values, I would run into an issue with any NAs/missingness in the columns. I think that it's currently enforced that we can't have emptiness for the allele columns in these file formats but I added [this parameter allow_na](https://github.com/Sage-Bionetworks/Genie/blob/gen-809-validate-allele/genie/validate.py#L460C35-L460C35) to flag any NAs as invalid otherwise the code will break when using `pandas.str.match`. We also have to have special handling to validate a column with all NA values in this function as well...

I talked more about the caveats of handling [NAs here](https://sagebionetworks.jira.com/browse/GEN-943)

**Testing:**
- Unit tests were written to test the general allele validation functions. 
- Maf and Vcf classes' tests were adjusted to account for the new allele validation. 
- Validation was run on the test genie project.

### DIFF
--- a/genie/validate.py
+++ b/genie/validate.py
@@ -424,7 +424,7 @@ def get_invalid_allele_rows(
     allowed_comb_alleles: list,
     allowed_ind_alleles: list,
     ignore_case: bool = False,
-    allow_na: bool = False,
+    allow_na: bool = False
 ) -> pd.Index:
     """
     Find invalid indices in a DataFrame column based on allowed allele values.
@@ -439,7 +439,6 @@ def get_invalid_allele_rows(
         ignore_case (bool, optional): whether to perform case-insensitive matching
         allow_na (bool, optional): whether to allow NAs to be an allowed allele
             value or not.
-
     Returns:
         pd.Index: A pandas index object indicating the row indices that
         don't match the allowed alleles
@@ -456,10 +455,23 @@ def get_invalid_allele_rows(
     else:
         flags = 0  # no flags
 
-    matching_indices = input_data[input_col].str.match(
-        search_str, flags=flags, na=allow_na
-    )
-    invalid_indices = input_data[~matching_indices].index
+    # special handling for all NA column
+    is_all_na = pd.isna(input_data[input_col]).all()
+    if is_all_na and allow_na:
+        invalid_indices = pd.Index([])
+    elif is_all_na and not allow_na:
+        invalid_indices = input_data.index
+    else:
+        # convert numeric cols to string while preserving NAs in order to use str.match
+        transformed_data = input_data.copy()
+        transformed_data[input_col] = transform._convert_col_with_nas_to_str(
+            transformed_data, input_col
+        )
+
+        matching_indices = transformed_data[input_col].str.match(
+            search_str, flags=flags, na=allow_na
+        )
+        invalid_indices = transformed_data[~matching_indices].index
     return invalid_indices
 
 
--- a/genie/validate.py
+++ b/genie/validate.py
@@ -424,7 +424,7 @@ def get_invalid_allele_rows(
     allowed_comb_alleles: list,
     allowed_ind_alleles: list,
     ignore_case: bool = False,
-    allow_na: bool = False
+    allow_na: bool = False,
 ) -> pd.Index:
     """
     Find invalid indices in a DataFrame column based on allowed allele values.
--- a/genie/validate.py
+++ b/genie/validate.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python3
+import re
 import logging
 from typing import Dict, List, Optional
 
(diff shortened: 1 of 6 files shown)


genie/validate.py
```diff
@@ -415,3 +416,61 @@ def standardize_string_for_validation(
...
+        input_data (pd.DataFrame): The DataFrame to search.
+        input_col (str): The name of the column to check.
+        allowed_alleles (list): The list of allowed allele values.
+        ignore_case (bool, optional): whether to perform case-insensitive matching
+
+    Returns:
+        pd.Index: A pandas index object indicating the row indices that
+        don't match the allowed alleles
+    """
+    search_str = rf"^[{''.join(allowed_alleles)}]+$"
```
thomasyu888: I think this regex does actually check if for alleles like 'ATATATATAT'

genie_registry/maf.py
```diff
@@ -70,6 +70,8 @@ class maf(FileTypeFormat):
     _fileType = "maf"
 
     _process_kwargs = []
+    _allele_cols = ["REFERENCE_ALLELE", "TUMOR_SEQ_ALLELE1", "TUMOR_SEQ_ALLELE2"]
+    _allowed_alleles = ["A", "T", "C", "G", "N", " ", "-"]
```
thomasyu888: I would remove " "  and "-".

I like what you're thinking here for class attributes.  In the future, these classes should follow SOLID principles, right now it is doing too much.  I can imagine having a `Validate` class that `ValidateMaf` inherits from and then different compositions if necessary.

For example, I can imagine having a `_required_columns` attribute.

genie/validate.py
```diff
@@ -415,3 +416,61 @@ def standardize_string_for_validation(
...
+    Returns:
+        tuple: The errors and warnings from the allele validation
+               Defaults to blank strings
+    """
+    errors = ""
+    warnings = ""
+    if len(invalid_indices) > 0:
+        errors = (
+            f"{fileformat}: Your {invalid_col} column has invalid allele values. "
+            f"These are the accepted allele values: {allowed_alleles}.\n"
```
thomasyu888: The alleles should be combinations of the allowed alleles.

tests/test_validate.py
```diff
@@ -768,3 +768,94 @@ def test_that_standardize_string_for_validation_returns_expected(
...
     )
     assert test_str == expected
+
+
+@pytest.mark.parametrize(
+    "input,expected_index,allowed_alleles,ignore_case",
+    [
+        (
+            pd.DataFrame(
+                {"REFERENCE_ALLELE": ["ACGT-G", "A-CGT ", "A", "C", "T", "G", "-", " "]}
```
thomasyu888:
```suggestion
                {"REFERENCE_ALLELE": ["ACGTG", "ACGT ", "A", "C", "T", "G", "-", " "]}
```

lets make the white space invalid.  Lets try and rely on pandas as much as possible, if pandas reads in the " "  as float('nan') then it's fine to include it here.

tests/test_validate.py
```diff
@@ -768,3 +768,94 @@ def test_that_standardize_string_for_validation_returns_expected(
...
+            True,
+        ),
+        (
+            pd.DataFrame({"REFERENCE_ALLELE": ["XXX", "ACGT"]}),
+            pd.Index([0]),
+            ["A", "T", "C", "G", " ", "-"],
+            True,
+        ),
+        (
+            pd.DataFrame({"REFERENCE_ALLELE": ["ACGT-G", pd.NA, None]}),
```
thomasyu888: Are `float('nan')` currently allowed in the maf?

genie_registry/vcf.py
```diff
@@ -18,6 +18,8 @@ class vcf(FileTypeFormat):
     _fileType = "vcf"
 
     _process_kwargs = []
+    _allele_col = "REF"
```
thomasyu888: We can just make it consistent and accept a list here.  That said, ALT is actually another allele column.  It's a little more complicated, so I think we should leave it out for now and see how genome nexus handles some of the files that aren't formatted properly

> ALT - alternate base(s): Comma separated list of alternate non-reference alleles. These alleles do not have to
be called in any of the samples. Options are base Strings made up of the bases A,C,G,T,N,*, (case insensitive)
or an angle-bracketed ID String (“<ID>”) or a breakend replacement string as described in the section on
breakends. The ‘*’ allele is reserved to indicate that the allele is missing due to a upstream deletion. If there
are no alternative alleles, then the missing value should be used. Tools processing VCF files are not required
to preserve case in the allele String, except for IDs, which are case sensitive. (String; no whitespace, commas,
or angle-brackets are permitted in the ID String itself)

genie/validate.py
rxu17: It does. Have to update it for '-', empty strings and NAs based on our discussion

genie_registry/maf.py
rxu17: Yes agreed, I originally had them defined in the `_validate` function and was debating between the two but was thinking future wise when we do want to start separating out into the separate classes we talked about (e.g Processing class, Validate class and regular read in data class), thought it would be easier if we put constants like these up in the class attributes given how big _validate is getting. Even before then, we might have other allele variable specific validation to add.

genie/validate.py
rxu17: Yes, message will be updated based on discussions earlier.

tests/test_validate.py
rxu17: Yes, thankfully `pandas.str.match` has a parameter that handles anything NA so we should be good on that front and don't need to add the long list of possible NA values here.

tests/test_validate.py
rxu17: According to [this code](https://github.com/Sage-Bionetworks/Genie/blob/develop/genie_registry/maf.py#L161-L171), it seems like it could be possible for allele columns to be read in as numeric and get something like `float('nan')` (i believe this is only for numeric columns). 

This makes me wonder if we should have standardization of columns and their datatypes before hand (either they have the wrong datatype and stop validation there, or they have the wrong datatype, and it gets converted before further validation) otherwise this just adds a lot more responsibilities for the validations that are string column specific as they will error out with numeric columns. 

Otherwise, will just have to add another if statement logic to check that the column is string or convert it to str because we won't be able to do our regex matching at all

genie_registry/vcf.py
rxu17: Yeah, I didn't think we would handle the alternative non-ref alleles. This looks like fun to handle if we did.

genie/validate.py
```diff
@@ -415,3 +416,96 @@ def standardize_string_for_validation(
...
+        search_str += f'|^[{re.escape("".join(allowed_ind_alleles))}]+$'
+
+    if ignore_case:
+        flags = re.IGNORECASE
+    else:
+        flags = 0  # no flags
+
+    # special handling for all NA column
+    is_all_na = pd.isna(input_data[input_col]).all()
+    if is_all_na and allow_na:
```
rxu17, line 460: Not the biggest fan of what has to be done here to allow the checking of non-str columns and columns with all NAs. I almost feel like it would be better to have the `_check_allele_col` be an enforced rule where it skips certain validation rules if the column has any NAs OR if we choose to allow NAs, have a standardized handling for those (if that's even possible...) so that we don't have to do it in the individual validation functions.

thomasyu888: Agreed, can you create a ticket to track that?  for now, let's do the 80/20 rule.  This does exactly what we need it to do.

thomasyu888 (approved):
🔥 I'm going to pre-approve - great work!  I'll look closer at the tests when I get home from the office.

thomasyu888 (commented):
👌👌 LGTM!

PR #540, opened 2023-11-16
[GEN-973] Fix bed file-clinical file cross-validation

rxu17:
**Purpose:** This is to add a fix to the cross validation for bed files against SEQ_ASSAY_IDs in clinical files

**Changes:** Changes were added to [_cross_validate_bed_files_exist](https://github.com/Sage-Bionetworks/Genie/blob/ab7e882a00e9c44cf2f5c51c6e7dd09e50afec22/genie_registry/clinical.py#L1015) to do the following:

- Split the original function into one that outputs the missing bed files and one that outputs errors/warnings for ease of testing
- [Standardizes the SEQ_ASSAY_IDs first then get a unique list of them to loop through](https://github.com/Sage-Bionetworks/Genie/blob/ab7e882a00e9c44cf2f5c51c6e7dd09e50afec22/genie_registry/clinical.py#L1021-L1024)

**Testing:**
- Tests were modified to have different case/underscored SEQ_ASSAY_ID values in the clinical file

### DIFF
--- a/genie_registry/clinical.py
+++ b/genie_registry/clinical.py
@@ -1032,7 +1032,7 @@ class Clinical(FileTypeFormat):
                 **exception_params,
             )
             if not bed_files["files"]:
-                missing_files.append(f"{seq_assay_id}.bed")
+                missing_files.append(f"{seq_assay_id.upper()}.bed")
         return missing_files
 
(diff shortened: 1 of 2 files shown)


thomasyu888 (approved):
🔥 Thanks for fixing this!  Nice find!

PR #541, opened 2023-11-28
[GEN-977] update annotation-tools version

rxu17:
**Purpose:** In order to support 'N' alleles, we had to update the annotation-tools repo and create a new release as a result. We will have to pin the new version in main genie.

Links to https://github.com/Sage-Bionetworks/annotation-tools/pull/7


thomasyu888 (approved):
💯 Great!

PR #542, opened 2023-12-11
Update docs

rxu17:
**Purpose:** Just a small PR to update documentation to be consistent with the logic of the pipeline.

**Changes Made**
- Updates to `README.md` to include the up to date logic in running the pipeline steps (Note that there is a JIRA ticket to discuss the current logic [here](https://sagebionetworks.jira.com/browse/GEN-1019))
- Updates to `CONTRIBUTING.md` to include section for running/testing docker, running pipeline steps on test pipeline

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -94,6 +94,10 @@ This package uses [semantic versioning](https://semver.org/) for releasing new v
 
 ### Testing
 
+#### Running test pipeline
+
+Make sure to run each of the [pipeline steps here](README.md#developing-locally) on the test pipeline and verify that your pipeline runs as expected. This is __not__ automatically run by Github Actions and have to be manually run.
+
 #### Running tests
 
 This package uses [`pytest`](https://pytest.org/en/latest/) to run tests. The test code is located in the [tests](./tests) subdirectory.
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -134,6 +138,17 @@ Follow gitflow best practices as linked above.
 1. Merge `main` back into `develop`
 1. Push `develop`
 
-### DockerHub
+### Modifying Docker
+
+Follow this section when modifying the [Dockerfile](https://github.com/Sage-Bionetworks/Genie/blob/main/Dockerfile):
+
+1. Make sure you have your synapse config setup in your working directory
```
thomasyu888: What is your rationale of doing this over using env variable to log into the python client?

thomasyu888 (approved):
🔥 nice! Just a small comment about using env variable to log into the client instead of the config file.

See the authentication portion of synapseclient docs.

CONTRIBUTING.md
rxu17: No rationale actually, I think I just carried this over from another project (specifying the env variable would be easier). Let me modify the docker command so it only has relevant flags in the command.

PR #544, opened 2024-01-05
[GEN-1026] Output annotation error report

rxu17:
**Purpose:** Part of the ticket to allow for the output of the annotation error report. 
This will add the new annotation-tools tag (once the annotation-tools PR gets merged and a new tag gets created) and the required java 11 update to the docker image in order to support the new `annotator.jar` file used as part of the GEN-1026 update.

**Testing:**
Rebuilt docker container in ec2 instance and ran the process mutation files. Files ran successfully.

Depends on [annotation-tools PR #8](https://github.com/Sage-Bionetworks/annotation-tools/pull/8)

PR #545, opened 2024-01-09
[GEN-1027] Store failed annotations error report

rxu17:
**Purpose:** This PR is intended to allow the concatenation and storage to Synapse of the failed annotations per center. This is currently a draft PR - there's some points of discussion needed in some areas.

**Changes:** 
**Old functions:**
- Split annotation_mutation into [create_annotation_paths](https://github.com/Sage-Bionetworks/Genie/blob/gen-1027-store-error-report/genie/process_mutation.py#L249) and revised [annotation_mutation](https://github.com/Sage-Bionetworks/Genie/blob/gen-1027-store-error-report/genie/process_mutation.py#L328) to improve testability, and ease of access to the filepaths used in the annotation

**New functions:**
These functions are created primarily for the concatenation and storage to synapse of the annotation reports. We also have a check function that just makes sure our narrow maf failed annotations count match the error report rows.
- [concat_annotation_error_reports](https://github.com/Sage-Bionetworks/Genie/blob/gen-1027-store-error-report/genie/process_mutation.py#L306)
- [check_annotation_error_reports](https://github.com/Sage-Bionetworks/Genie/blob/gen-1027-store-error-report/genie/process_mutation.py#L333)
- [store_annotation_error_reports](https://github.com/Sage-Bionetworks/Genie/blob/gen-1027-store-error-report/genie/process_mutation.py#L360)

**Testing:**
- Unit and integration tests added to, to account for the new and revised functions above
- Ran on docker image built on this branch
- Ran processing on just mutation data and files show up here in the **Errors** folders for each center: https://www.synapse.org/#!Synapse:syn11601248/tables/

Depends on #544


tests/conftest.py
```diff
@@ -51,13 +51,23 @@ def genie_config():
...
                 "stagingSynId": "syn11601342",
+                "errorsSynId": "syn53239081",
+                "release": True,
+                "mutationInCisFilter": "ON",
+            },
+            "GOLD": {
+                "center": "GOLD",
+                "inputSynId": "syn52950860",
+                "stagingSynId": "syn52950861",
+                "errorsSynId": "syn53239077",
```
rxu17, line 70: I am hesitant to follow our original plan of creating this nested file structure on synapse: `<center_name>/errors/processing/mutation/failed_annotations_report.tsv` vs just placing the failed annotations in `<center_name>/errors/failed_annotations_report.tsv` just because:

1. We don't have many error reports and don't forsee us having many distinct ones - we have the validation errors, and now the mutation annotation errors
2. We'd have to create that nested structure across all centers (would probably have to automate it otherwise that would take some time
3. increased complexity for testing

Thoughts?

genie/process_mutation.py
```diff
@@ -226,18 +227,110 @@ def process_mutation_workflow(
...
+    full_error_report = pd.concat(error_reports)
+    full_error_report["Center"] = center
+    return full_error_report
+
+
+def store_annotation_error_reports(
+    full_error_report: pd.DataFrame,
+    syn: Synapse,
+    errors_folder_synid: str,
+    release: str,
```
rxu17: I don't see anything related to a processing run version comment as part the parameters available for running `input_to_database`. 

That only occurs with running the consortium release and so forth. I'm guessing what we probably want here for consistency with our other files is for when we run consortium release / public release code to have a function there to comment on the version of the annotation error report (I'm guessing we will always pull the latest version) and mark it as 15.3 consortium release for example? Over trying to add that in during processing here.

tests/conftest.py
thomasyu888: I agree with this assessment.  Let;'s just try and keep it simple - we can always make the filename more informative

genie/process_mutation.py
thomasyu888: Hmm you do make a good point.  I think once we get around to separating validation and processing, we will be able to split up the `release` parameter better.  I think it would make this PR overly complex if we were to add in the `release` parameter here.  We do keep track of when the processing code is executed, so the error reports are just updated whenever the processing is run.

To make validation + processing truly idempotent, ideally we are passing in a "manifest" of entities with it's versions as the input.  I did have a question: what happens when there are no annotation errors, is an empty file with headers uploaded?

genie/process_mutation.py
```diff
@@ -226,18 +227,110 @@ def process_mutation_workflow(
...
+    store_annotation_error_reports(
+        full_error_report=full_error_report,
+        syn=syn,
+        errors_folder_synid=genie_config["center_config"][center]["errorsSynId"],
+        release="test",
+    )
+    return annotation_paths.merged_maf_path
+
+
+def create_annotation_paths(center: str, workdir: str) -> namedtuple:
```
thomasyu888, line 256: Nice use of `namedtuple` here - I personally think this is a great example of when to use something like this over a `class` to define the structure of `Filepaths`.  That said - nothing wrong with classes either.

What was your rationale in picking `namedtuple` or `class`?

genie/process_mutation.py
```diff
@@ -226,18 +227,110 @@ def process_mutation_workflow(
...
+    full_error_report = concat_annotation_error_reports(
+        center=center,
+        input_dir=annotation_paths.error_dir,
+    )
+
+    store_annotation_error_reports(
+        full_error_report=full_error_report,
+        syn=syn,
+        errors_folder_synid=genie_config["center_config"][center]["errorsSynId"],
+        release="test",
```
thomasyu888: As per comment above, I'd remove this.

genie/process_mutation.py
```diff
@@ -226,18 +227,110 @@ def process_mutation_workflow(
...
+    chunk_size = 10000
+    error_reports = []
+
+    # Read and concatenate TSV files in chunks
+    for file in error_files:
+        for chunk in pd.read_csv(
+            os.path.join(input_dir, file), sep="\t", chunksize=chunk_size
+        ):
+            error_reports.append(chunk)
+    full_error_report = pd.concat(error_reports)
```
thomasyu888, line 328: Nit: Querying the mutation table shows that there are ~30000 failed annotations.  Let's not prematurely optimize, but I just wanted to point out that this _could be_ an issue if there ends up being lots of annotation failures for some reason.

thomasyu888 (changes requested):
Nice - just one small change, but this is looking great!

genie/process_mutation.py
rxu17: Sounds good to the `release` parameter + future validation + processing thoughts.


> what happens when there are no annotation errors, is an empty file with headers uploaded?

Yes, when there are no annotation errors, it's just an empty file with headers uploaded (gn tool does this and I just decided to keep it consistent with them). Is this preferred behavior? Given that we will have a summary table telling sites if they have any annotation errors at all.

genie/process_mutation.py
rxu17: Gotcha! Agreed, I was just taking precautions given that we're optimizing for when we create the compiled annotated maf file. Will remove for now.

genie/process_mutation.py
rxu17: I picked `namedtuple` mainly because of its immutability and simplistic, readable structure since we have the fixed set of filepaths we plan to use and we don't plan to change (or need any additional functionality as far as i can forsee) those filepaths in the course of the code. And we don't need to access those filepaths except for this portion of the code.

genie/process_mutation.py
thomasyu888: @rxu17 no need to remove the chunking. I'm saying there might be a day when `pd.concat` may be the issue.

Sorry I wasn't clear!

rxu17: Makes sense, appending/concatenating the chunks as it goes along into a dataframe seems more memory efficient than pd.concat at the end with everything at once. Then there is the side of it being possibly slower than pd.concat. But yes, something to investigate when we do get into noticeably larger annotation failures. Will put that in a confluence page for now as we have areas of the pipeline we could optimize better. 

Added it here: [6.1 Memory-profiling](https://sagebionetworks.jira.com/wiki/spaces/APGD/pages/3024650251/Main+GENIE+Developer+Title+Page#6.1-Memory-profiling)

genie/process_mutation.py
thomasyu888: That is ok.  No harm in uploading an empty file.

genie/process_mutation.py
thomasyu888: Agreed!

thomasyu888 (approved):
🔥 LGTM! I'm going to pre-approve!

genie/process_mutation.py
```diff
@@ -226,18 +230,163 @@ def process_mutation_workflow(
...
+    """
+    maf_table_df = extract.get_syntabledf(
+        syn=syn,
+        query_string=(
+            f"SELECT * FROM {maf_table_synid} "
+            f"WHERE Center = '{center}' AND "
+            "Annotation_Status = 'FAILED'"
+        ),
+    )
+    assert len(maf_table_df) == len(full_error_report), (
```
thomasyu888: @rxu17 this is a great idea, but also a tiny bit risky for processing itself because mutation processing takes days.  I wonder if we want to log this instead of throwing an error that would pause processing all together

rxu17:
Not adding it in for this PR, but given that we don't want to enforce most things as errors unless it's absolutely necessary during processing, probably want to play around with adding custom warning classes. Also easier for categorizing and filtering out warnings.

genie/process_mutation.py
rxu17: @thomasyu888 Sounds good, changed to a warning. 

Not adding it in for this PR, but given that we don't want to enforce most things as errors unless it's absolutely necessary during processing, probably want to play around with adding custom warning classes. Makes it easier for categorizing and filtering out warnings.

thomasyu888: @rxu17 can you create a ticket around that work?

rxu17: Yes created here: [GEN-1081](https://sagebionetworks.jira.com/browse/GEN-1081)

rxu17:
Note that this PR introduces a new variable `genomic_location_explanation` to the full maf file during mutation processing that leads an issue with the `store_maf_files` function in `genie/database_to_staging.py` that will be resolved with #548.  See that PR for more details.

PR #547, opened 2024-01-15
[GEN-1078] HOTFIX: Add in new known string columns as MSK added new columns into maf file

thomasyu888:
Note: This PR is being merged into main to enable processing.

**Problem**

The first 5000 rows of file do not always contain the correct dtype

**Solution**
Set the known string columns to be string type so that the code doesn't try to convert string to integer


Chelsea-Na (approved):
If possible we should probably explore how we can ignore casing on column headers when we’re reading in the data frame (or an alternative solution) so we don’t face the “transcription” or “TRANSCRIPTION” issue. 

We should also discuss if it is possible to only read in specific columns when reading in the data frame in pandas.

PR #548, opened 2024-01-24
[GEN-1076] Exclude genomic_location_explanation from release

rxu17:
**Purpose:** This is a draft PR. This PR is only allowing in certain release variables (excluding out the `genomic_location_explanation` and `Annotation_Status` variables for example) for the full maf csv.

**Changes:**
- Added `create_missing_columns` in `process_functions.py` Decided with this implementation on excluding variables because we have older full maf files that may have been processed 1-N releases ago, and doesn't contain new columns and features we added to maf processing since then. Because we append old full maf files and more recently processed maf files, we want to make sure they conform to a specific expected schema so we set the expected columns they should have and if they don't have the column, it's created and filled with missing values.

**Testing:**
- Unit tests written for the new function
- Ran on consortium and public release aspects of pipeline
- Ran in docker image

Depends on #545


genie/database_to_staging.py
```diff
@@ -35,6 +35,74 @@
 SV_CENTER_PATH = os.path.join(GENIE_RELEASE_DIR, "data_sv_%s.txt")
 BED_DIFFS_SEQASSAY_PATH = os.path.join(GENIE_RELEASE_DIR, "diff_%s.csv")
 
+FULL_MAF_RELEASE_SCHEMA = {
```
rxu17: One outstanding feature to this PR is if should we have a function that enforces these expected data types on the release file? Right now we just assume they are the expected / I pulled this from one of the full maf release files in the test pipeline and got their data types.

genie/process_functions.py
```diff
@@ -948,3 +948,35 @@ def create_new_fileformat_table(
...
+    for column, data_type in schema.items():
+        if column not in dataset.columns:
+            dataset = dataset.assign(**{column: missing_values[data_type]})
+
+        # only way to preserve NAs for these specific dtype columns
+        if data_type == "integer":
+            dataset[column] = dataset[column].astype("Int64")
+        elif data_type == "boolean":
+            dataset[column] = dataset[column].astype(pd.BooleanDtype())
+    return dataset[list(schema.keys())]
```
rxu17, line 982: On the fence about having this function also only including expected columns based on the input schema, given the function is `create_missing_column` but at the same time, since we're adding missing columns because it's not following the expected schema, maybe it does make sense for it to just subset the dataset on the expected columns

genie/process_functions.py
```diff
@@ -948,3 +948,35 @@ def create_new_fileformat_table(
...
+        "string": "",
+        "integer": None,
+        "float": float("nan"),
+        "boolean": None,
+    }
+    for column, data_type in schema.items():
+        if column not in dataset.columns:
+            dataset = dataset.assign(**{column: missing_values[data_type]})
+
+        # only way to preserve NAs for these specific dtype columns
```
rxu17, line 977: One behavior yet to be tested in `database_to_staging` is what are the downstream effects of converting to these pandas nullable data types (data types like int and boolean that typically doesn't allow NAs but these ones were created to allow NAs). 

But at the same time, we also don't know the data types of previously processed files vs newly processed files because they could be using different data types, so this helps conform that. Maybe another reason to enforce the schema data types. It doesn't seem like we've run into problems thus far with appending full maf files that were old and new for release but ... 

I was tempted to just omit this function and have us just subset the dataset to a list of expected full maf release variables and call that done as that's the simplest solution (with least downstream effects to test) and we aren't allowing in any new release variables so technically production data should still run given that before we just pulled the header of the 1st dataset on the file view list and used that as our expected columns designated.

I think all of this is also just further indication of needing more "realistic" test data to be more confident in the above implementation.

rxu17:
Also feels like we need to also enforce an expected schema for the full maf file during the [end of process_mutation](https://github.com/Sage-Bionetworks/Genie/blob/gen-1076-exclude-new-gn-var/genie/process_mutation.py#L486) given that we have this new variable `genomic_location_explanation`.

genie/database_to_staging.py
thomasyu888: Hmm, can you speak to how we would enforce these data types?  We do read the maf dataframe in with these dtypes right?

genie/process_functions.py
thomasyu888: I agree, I'm happy for you to comment out this function since you wrote it and revisit, but I think we should have a set schema for the MAF.

Definitely look at the simplest solution with least amount of tech debt.

genie/database_to_staging.py
rxu17: Yeah it seems like the maf dataframe gets read in and it takes these dtypes. We don't have code in the read_csv that specifies the dtype by column, it's just a regular [read_csv](https://github.com/Sage-Bionetworks/Genie/blob/develop/genie/database_to_staging.py#L769-L771)

We could enforce it when we read it in like this for example: `df = pd.read_csv(maf_data, dtype=FULL_MAF_RELEASE_SCHEMA)`

genie/process_functions.py
rxu17: Won't use

rxu17:
It was decided to just use a list of release variables for the full maf file to subset out the non-release variables.

genie/database_to_staging.py
```diff
@@ -35,6 +35,73 @@
 SV_CENTER_PATH = os.path.join(GENIE_RELEASE_DIR, "data_sv_%s.txt")
 BED_DIFFS_SEQASSAY_PATH = os.path.join(GENIE_RELEASE_DIR, "diff_%s.csv")
 
+FULL_MAF_RELEASE_COLUMNS = [
```
thomasyu888, line 38: Are these all the columns in the release maf right now?

thomasyu888 (approved):
🚀 LGTM!  Thanks for doing this!  I just had one comment, but going to pre-approve!

genie/database_to_staging.py
rxu17: Yeah, I looked at the [AACR maf file columns from the 16.1 consortium release](https://www.synapse.org/#!Synapse:syn5571527) and used those columns. Hopefully that's representative of all of the release maf columns we would have

PR #549, opened 2024-02-06
[GEN-974] Allow NaN, nan and NA strings for mutation data

rxu17:
**Purpose:** This is a draft PR. This PR allows in the "NaN", "nan" and "NA" strings for allele columns in the vcf and maf datasets because these are valid allele combinations.

**Changes:**
- [_convert_values_to_na in genie/transform.py](https://github.com/Sage-Bionetworks/Genie/pull/549/files#diff-f5e09cd8a62285bf7d3ff1917b948de18ec663c78a15a600176178682757fd8eR70)- new function to convert all occurrences of values in a dataframe to NA, this is a helper function that is used in the `_get_dataframe` methods of the `vcf.py` and `maf.py` files
- [_get_dataframe in maf.py](https://github.com/Sage-Bionetworks/Genie/pull/549/files#diff-323044b86c877fd4af48462d3e131927c1a76d6ecef14abf031d1847a7619718L373-L403)
- [_get_dataframe in vcf.py](https://github.com/Sage-Bionetworks/Genie/pull/549/files#diff-f5006d44639bc1a25df2986607f6edaf6f310b9139ef670420d09a38f5e04c98L40-L45)
-  I decided to add the allowing in of the "NaN", "nan" and "NA" strings in this method because we need this to occur for validation and processing, and this is the method that is used in both. I also had to do some special handling for maf files because they can allow in case insensitive column names. The in-depth reasoning behind allowing in  "NaN", "nan" and "NA" strings for ALL of the dataset, then converting them back to NA in **non-allele columns** can be found in the comments of the [JIRA ticket](https://sagebionetworks.jira.com/browse/GEN-974). Summary of reasoning is that it is difficult to use`read_csv` to only convert specific columns, especially since it already has a bunch of arguments for NA specific handling, and order of operations comes into play here. See the [docs for more details](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html#pandas-read-csv). 

**Testing:**
- [X] Updated and ran pytests locally
- [X] Updated maf and vcf test files in test synapse project to include NaN, nan and NA as valid allele values and ran genie pipeline all the way through locally (integration test)
- [X] Ran genie pipeline under docker image

### DIFF
--- a/genie/transform.py
+++ b/genie/transform.py
@@ -1,6 +1,7 @@
 """This module contains all the transformation functions used throughout the GENIE
 package"""
 
+from typing import List
 import warnings
 
 import pandas as pd
@@ -68,14 +69,14 @@ def _convert_df_with_mixed_dtypes(read_csv_params: dict) -> pd.DataFrame:
 
 
 def _convert_values_to_na(
-    input_df: pd.DataFrame, values_to_replace: list, columns_to_convert: list
+    input_df: pd.DataFrame, values_to_replace: List[str], columns_to_convert: List[str]
 ) -> pd.DataFrame:
     """Converts given values to NA in an input dataset
 
     Args:
         input_df (pd.DataFrame): input dataset
-        values_to_replace (list): string values to replace with na
-        columns_to_convert (list): subset of columns to convert with na in
+        values_to_replace (List[str]): string values to replace with na
+        columns_to_convert (List[str]): subset of columns to convert with na in
 
     Returns:
         pd.DataFrame: dataset with specified values replaced with NAs
--- a/genie/transform.py
+++ b/genie/transform.py
@@ -64,3 +64,24 @@ def _convert_df_with_mixed_dtypes(read_csv_params: dict) -> pd.DataFrame:
         df = pd.read_csv(**read_csv_params, low_memory=False, engine="c")
     warnings.resetwarnings()
     return df
+
+
+def _convert_values_to_na(
+    input_df: pd.DataFrame, values_to_replace: list, columns_to_convert: list
+) -> pd.DataFrame:
+    """Converts given values to NA in an input dataset
+
+    Args:
+        input_df (pd.DataFrame): input dataset
+        values_to_replace (list): string values to replace with na
+        columns_to_convert (list): subset of columns to convert with na in
+
+    Returns:
+        pd.DataFrame: dataset with specified values replaced with NAs
+    """
+    if not input_df.empty:
+        replace_mapping = {value: None for value in values_to_replace}
+        input_df[columns_to_convert] = input_df[columns_to_convert].replace(
+            replace_mapping
+        )
+    return input_df
--- a/genie/transform.py
+++ b/genie/transform.py
@@ -1,5 +1,6 @@
 """This module contains all the transformation functions used throughout the GENIE
 package"""
+
 import warnings
 
 import pandas as pd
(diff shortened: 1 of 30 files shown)


genie/transform.py
```diff
@@ -64,3 +65,24 @@ def _convert_df_with_mixed_dtypes(read_csv_params: dict) -> pd.DataFrame:
...
+    Args:
+        input_df (pd.DataFrame): input dataset
+        values_to_replace (list): string values to replace with na
+        columns_to_convert (list): subset of columns to convert with na in
+
+    Returns:
+        pd.DataFrame: dataset with specified values replaced with NAs
+    """
+    if not input_df.empty:
+        replace_mapping = {value: None for value in values_to_replace}
```
thomasyu888, line 85: I see you're converting to `none` here, but should it be `float('nan')`

genie_registry/maf.py
```diff
@@ -370,34 +396,43 @@ def _get_dataframe(self, filePathList):
...
             # with R, quote=T
             "quoting": 3,
             # Retain completely blank lines so that
             # validator will cause the file to fail
             "skip_blank_lines": False,
         }
+
         mutationdf = transform._convert_df_with_mixed_dtypes(read_csv_params)
+
+        mutationdf = transform._convert_values_to_na(
```
thomasyu888, line 425: 👀 I did not think about this at all... Good catch!

thomasyu888 (approved):
🔥 Fantastic!  Going to approve, just a small comment.

genie/transform.py
rxu17: Thanks for bringing this up.

So, I'm seeing [this in maf.py's _validation](https://github.com/Sage-Bionetworks/Genie/pull/549/files#diff-323044b86c877fd4af48462d3e131927c1a76d6ecef14abf031d1847a7619718L199-R214) in which this is handled (originally). I'm not seeing anything in processing in annotation-tools for maf.py that resembles what is happening in validation. Which is surprising, because I would think we'd want to do the same?

I think I have to move part of the code from the above specifically to the [_convert_values_to_na in genie/transform.py](https://github.com/Sage-Bionetworks/Genie/pull/549/files#diff-f5e09cd8a62285bf7d3ff1917b948de18ec663c78a15a600176178682757fd8eR70) method just to make sure it's int/float dtypes for both validation and processing

Or am I missing something and we don't need them to be numeric for processing?

rxu17: After experimenting, I think having `None` works. A dataframe with string values and None (e.g: `pd.DataFrame({"bye":["2314", None, "124"]})`) converted to float will convert the column to float just fine.

rxu17: It seems like for `vcf` files' validation, we don't have to worry about this numeric stuff handling in validation or processing because `POS` is the only expected numeric column and it can't have NAs anyways.

So just maf processing side needs the handling

thomasyu888: So most of the heavy lifting of processing is done via genome nexus and once we write it out into a csv, it doesn't matter what "type" the column is. 

That said - we do want to be diligent about making sure the data itself isn't changing, but a '1' and 1 is going to look the same when we write it out (unless we deliberately add quotes)

rxu17: Ah I see, then I think it's okay as it is? Since we've come this far without much issue without having to implement that for the maf processing side. I am for keeping `None` mainly because we can have more than just numeric columns and conversion to numeric works fine even with `None` values in the column

PR #552, opened 2024-03-09
[GEN-1191] Add file logging

rxu17:
**Purpose:** This small PR adds specific file logging when getting all of the center input files for validation/processing. See JIRA ticket for more details but in the past we had run into issues during this step and knowing the exact file prior to the retrieving of the file is helpful. 

**Testing:**
- Tested locally
- Tested on nextflow workflow

### DIFF
--- a/genie/extract.py
+++ b/genie/extract.py
@@ -13,8 +13,10 @@ import pandas as pd
 
 from genie import process_functions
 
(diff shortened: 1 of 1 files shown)


danlu1 (approved):
LGTM

PR #553, opened 2024-03-09
[GEN-1067] add functionality to warn for identical ref and tsa2

danlu1:
Purpose: 
This PR includes changes in `_check_tsa1_tsa2` function of `maf.py` to throw errors (column level and row level) when REFERENCE_ALLELE is equal to TUMOR_SEQ_ALLELE2. 

Test:
1. Unit tests are included. 
2. Updated  [GOLD center file](https://www.synapse.org/#!Synapse:syn52955202) and [Test center file](https://www.synapse.org/#!Synapse:syn51611938) and ran validation locally and on ec2 instance.

### DIFF
--- a/genie_registry/maf.py
+++ b/genie_registry/maf.py
@@ -29,9 +29,18 @@ def _check_tsa1_tsa2(df):
                 "All values in TUMOR_SEQ_ALLELE1 must match all values in "
                 "REFERENCE_ALLELE or all values in TUMOR_SEQ_ALLELE2.\n"
             )
-    if tsa2_col_exist and ref_col_exist and not df.query('REFERENCE_ALLELE == TUMOR_SEQ_ALLELE2').empty:
-        error =(f"{error}REFERENCE_ALLELE should not equal to TUMOR_SEQ_ALLELE2. Please check row: {', '.join(str(e+1) for e in df.query('REFERENCE_ALLELE == TUMOR_SEQ_ALLELE2').index.values)}.\n")
+    if (
+        tsa2_col_exist
+        and ref_col_exist
+        and not df.query("REFERENCE_ALLELE == TUMOR_SEQ_ALLELE2").empty
+    ):
+        error = (
+            "maf: Contains instances where values in REFERENCE_ALLELE match values in TUMOR_SEQ_ALLELE2. "
+            "This is invalid. Please correct."
+        )
+        row_index = df.query("REFERENCE_ALLELE == TUMOR_SEQ_ALLELE2").index.values
     return error
+
 
 def _check_allele_col(df, col):
     """
--- a/genie_registry/maf.py
+++ b/genie_registry/maf.py
@@ -11,7 +11,7 @@ from genie import process_functions, transform, validate
 logger = logging.getLogger(__name__)
 
 
-def _check_tsa1_tsa2(df):
+def _check_allele_col_validity(df):
     """If maf file has both TSA1 and TSA2,
     TSA1 must equal REF, or TSA1 must equal TSA2, and REF must not equal TSA2
     """
@@ -270,7 +270,7 @@ class maf(FileTypeFormat):
         #             "start with 'chr' or any 'WT' values.\n"
         #         )
 
-        error = _check_tsa1_tsa2(mutationDF)
+        error = _check_allele_col_validity(mutationDF)
         total_error.write(error)
 
         if process_functions.checkColExist(mutationDF, "TUMOR_SAMPLE_BARCODE"):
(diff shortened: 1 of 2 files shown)


genie_registry/maf.py
```diff
@@ -29,6 +29,9 @@ def _check_tsa1_tsa2(df):
                 "All values in TUMOR_SEQ_ALLELE1 must match all values in "
                 "REFERENCE_ALLELE or all values in TUMOR_SEQ_ALLELE2.\n"
             )
+    if tsa2_col_exist and ref_col_exist and not df.query('REFERENCE_ALLELE == TUMOR_SEQ_ALLELE2').empty:
+        error =(f"{error}REFERENCE_ALLELE should not equal to TUMOR_SEQ_ALLELE2. " 
+                f"Please check row: {', '.join(str(e+1) for e in df.query('REFERENCE_ALLELE == TUMOR_SEQ_ALLELE2').index.values)}.\n")
```
rxu17: So I wouldn't add in the row-based validation in the error message yet just because we haven't decided to start implementing and delivering row-based validation for sites as a whole and we would have to talk about the standard for doing that so we are consistent across all validation rules.

Since you already have code for the row-based validation, I would just remove it from the validation message but save it as a variable.

rxu17: Could you update the error message to mirror the structure of the previous error message in this function? Something like:
"maf: Contains instances where values in REFERENCE_ALLELE match values in TUMOR_SEQ_ALLELE2. This is invalid. Please correct."

tests/test_maf.py
```diff
@@ -258,6 +258,39 @@ def test_invalid__check_tsa1_tsa2():
         "REFERENCE_ALLELE or all values in TUMOR_SEQ_ALLELE2.\n"
     )
 
+def test_invalid__check_ref_tsa2():
```
rxu17: Could you add the following test cases (because of the various combos of columns the maf file can have):
- You don't have one or more of the columns required to perform a comparison and you get **no** error messages
- You don't have the `TUMOR_SEQ_ALLELE1` column and you get an error message
- You don't have the `TUMOR_SEQ_ALLELE1` column and you get **no** error message
- You have all three columns and you get **no** error message

rxu17 (commented):
Great first genie PR! I added some comments

rxu17:
Also don't forget to make sure your build is successful, i think there's just some formatting to adjust in your code: `https://github.com/Sage-Bionetworks/Genie/actions/runs/8237606927/job/22526728071?pr=553`

If you have `black` installed, you can just run black on everything using `black .` (which will auto-lint your code) then commit and push the formatting changes.

genie_registry/maf.py
danlu1: Done.

tests/test_maf.py
danlu1: Thanks so much for the comments. QQ: do you mean __tsa2__ in the second bullet point? Since tsa1 is optional, we don't throw an error when it's absent. Also, I think the fourth case has been checked in `test_valid__check_tsa1_tsa2()` already. Do I still need to add a test for it?

rxu17: Ah right, I forgot that there's `test_valid__check_tsa1_tsa2` and `test_invalid__check_tsa1_tsa2`. I wonder then if it might be better to change your function name to `_check_ref_tsa1_tsa2` because you're checking all three. And just add your relevant test cases to `test_valid__check_tsa1_tsa2` and `test_invalid__check_tsa1_tsa2` (and rename the test functions?).

genie_registry/maf.py
```diff
@@ -13,7 +13,7 @@
 
 def _check_tsa1_tsa2(df):
```
rxu17: NIT: Could you change the title of the function  (and the test function name) to be something more informative like `_check_allele_col_relationships`?

tests/test_maf.py
```diff
@@ -241,47 +241,110 @@ def test_error__check_allele_col():
...
+    ids=[
+        "matching_tsa1_tsa2",
+        "matching_tsa1_ref",
+        "invalid_tsa1",
+        "identical_ref_tsa2",
+        "identical_ref_tsa2_missing_tsa1",
+        "valid_ref_tsa2_missing_tsa1",
+        "missing_tsa2_ref",
+        "invalid_tsa1_identical_ref_tsa2",
+    ],
```
rxu17, line 358: Could you add a test case where we have NAs just to make sure the code doesn't break with NAs
- one test case of df with ref and tsa2 and tsa1 columns with NAs sprinkled in

rxu17 (approved):
Looks great, thanks for implementing the changes. Going to pre-approve! Just a couple more comments.

genie_registry/maf.py
danlu1: Updated function name to `_check_allele_col_validity`

tests/test_maf.py
danlu1: Sure. On it.

PR #554, opened 2024-03-14
[GEN-996] convert the clinical template columns from a set to a list

danlu1:
Problem:
Run into ValueError: columns cannot be a set when upgrading to pandas >= 1.5 for `Clinical.preprocess`

Solution:
Convert set to list

Test:
unit test function has been added

PR #555, opened 2024-03-15
[GEN-997] update method to concatenate columns for dataframe

danlu1:
Problem:
Cannot generate UNIQUE_KEY column by concatenating primary_cols for empty dataframe

Solution:
Special handling for empty dataframes: have them create the column directly and set it to ““

Test:
unit test function has been added

Depends on #554

### DIFF
--- a/tests/test_load.py
+++ b/tests/test_load.py
@@ -1,12 +1,13 @@
-from unittest.mock import patch, mock_open
-
+from unittest.mock import patch, mock_open, create_autospec
+import unittest
 import synapseclient
 from synapseclient.core.exceptions import SynapseTimeoutError
 
-from genie import load, __version__
+from genie import load, __version__, process_functions
 import pandas as pd
 import tempfile
 import os
+import pytest
 
 
 def test_store_file(syn):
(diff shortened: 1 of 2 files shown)


tests/test_load.py
```diff
@@ -67,3 +70,96 @@ def test_store_table_error(syn):
     with patch.object(syn, "store", side_effect=SynapseTimeoutError) as patch_store:
         load.store_table(syn, "full/path", "syn1234")
         patch_store.assert_called_once()
+
+
+def test__update_table_non_empty_dataframe(syn):
+    """Test _update_table function with both new_dataset and database as non_empty dataframe"""
+    database_synid = "syn123"
+    primary_key_cols = ["test", "foo"]
+    to_delete = (False,)
```
rxu17: Because the `_update_table` function has `to_delete` as a bool, it would be good to have a test case where `to_delete` is True and you can mock `process_functions._delete_rows` and call `assert_called_once_with` on it so you can make sure it's called with the expected arguments

tests/test_load.py
rxu17, line 1: Since you're writing tests for the entirety of the function `_update_table`, I would recommend that you separate out the `_update_table` into multiple functions so you can focus on having test cases for the part of the function that you changed otherwise it will be hard to add to these test cases in the future and also be difficult to troubleshoot failing test cases. This way you can also separate out the functionality of the various things in this function and test them independently from each other (and that they should be working independently from each other).

rxu17, line 1: Does this code work for empty DataFrames like `pd.DataFrame()`?

rxu17 (commented):
Thanks for this! Had a few comments

tests/test_load.py
danlu1: @rxu17 yes, it's better if we separate `_update_table` into different functions. I will work on it.

danlu1: It should. I will add test case for this after refactoring `_update_table`

tests/test_load.py
danlu1: Will check on this. Thanks for catching this.

thomasyu888 (approved):
🔥 LGTM!  Thanks for modularizing the code and for the great review!

PR #556, opened 2024-04-02
[GEN-1164] add workflow to build docker image  and push it to ghcr

danlu1:
Problem:

There are limited admin seats that we can have on dockerHub.

Solution:
Switch to pushing our docker images to ghcr

Output:

The docker image and cached image can be viewed [here](https://github.com/Sage-Bionetworks/Genie/pkgs/container/genie/versions).


.github/workflows/docker_build.yml
thomasyu888, line 1: @danlu1 this is great, but can we add this to our existing github action?  We don't necessarily want to build if the tests fail

.github/workflows/docker_build.yml
```diff
@@ -0,0 +1,53 @@
...
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Extract Docker metadata
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
```
rxu17: Could you add the following under `images:`
```
tags: |
    type=semver,pattern={{version}}
    type=ref,event=branch
    ref=main
    ref=develop
    type=sha
    latest
```
This will make sure we have docker image for the following tags `develop`, `main`, `latest`, and version of the tag I don't know if we need to have docker images created and tagged per feature branch as well.

.github/workflows/docker_build.yml
danlu1: Okay. I will work on this.

.github/workflows/docker_build.yml
danlu1: @rxu17 Thx for the comments. Could you plz point me to the doc describing using `ref=` in the tags sections? My original thoughts on achieving this goal are adding `main` branch to the push event and only use     
```
type=semver,pattern={{version}}
type=ref,event=branch
type=ref,event=tag
type=raw,value=latest,enable={{is_default_branch}} # to conditionally tag with latest for the default branch
``` 
for tags section.

danlu1: Based on our discussion on slack, we would like to add docker build step in `ci` workflow, create a docker image for develop, tag, and feature branch, and to use Registry cache since gha has a 10GB per repo storage limit.

.github/workflows/ci.yml
```diff
@@ -5,15 +5,18 @@ name: build
 
 on:
   push:
-    branches:
-      - develop
+    branches: [develop, 'GEN*', 'gen*']
```
rxu17: Just a comment to note that `test` and `lint` will run for every push now to the feature branches regardless of whether there is a PR or not. Nothing to fix here.

rxu17: Actually could you also add `main` to the branches here? I don't know if your current setup for ghcr will always create a docker image that is based on the latest version of the repo like we have for dockerhub? Will this setup create a `latest` tag by default?

rxu17 (commented):
Thanks for tackling this! Just a question

.github/workflows/ci.yml
danlu1: Yes, right now, push to develop, branches with gen/GEN prefix, all PRs and release will trigger the test and lint job.

rxu17 (approved):
LGTM!

thomasyu888 (commented):
🔥 thanks for the great work!  Great review

PR #558, opened 2024-04-16
[GEN-1000] Refactor and sort unmapped oncotree codes in validation message

rxu17:
**Purpose:** This is a draft PR. This is to handle the different clinical validation messages we get when running oncotree code validation. I wasn't able to replicate the discrepancy that I saw previously, so my resolution is to sort the codes we output so we don't run into said issue. This is a minor discrepancy and won't affect data quality that other issues when converting to pandas 2.0 may affect

Previous behavior I noticed:
- pandas 1.4 behavior
Validation message is “Please double check that all your ONCOTREE CODES exist in the mapping. You have 2 samples that don't map. These are the codes that don't map: XXXX,TMT”
	
- pandas 2.0 behavior
Validation message is “Please double check that all your ONCOTREE CODES exist in the mapping. You have 2 samples that don't map. These are the codes that don't map: TMT,XXXX”

**Changes:**
- Refactored the code in `_validate` in `clinical.py` that validated the unmapped oncotree codes into two validation functions - [one for outputting the indices in the input clinical file with the unmapped codes and one for outputting the error message](https://github.com/Sage-Bionetworks/Genie/blob/gen-1000-set-autosort/genie_registry/clinical.py#L475-L541). This is to set us up for row-based validation and having each validation be in its own independent function
- Have the [unmapped codes be sorted ](https://github.com/Sage-Bionetworks/Genie/blob/gen-1000-set-autosort/genie_registry/clinical.py#L530) so we avoid having different order of the codes

**Testing:**
- Wrote unit tests
- Tested validation pipeline with updates locally and on Tower using [newly built docker image here](https://github.com/Sage-Bionetworks/Genie/pkgs/container/genie/204549119?tag=gen-1000-set-autosort)

Depends on #555

### DIFF
--- a/genie_registry/clinical.py
+++ b/genie_registry/clinical.py
@@ -498,13 +498,12 @@ class Clinical(FileTypeFormat):
         ]
         return unmapped_oncotrees.index
 
-
     def _validate_oncotree_code_mapping_message(
         self: "Clinical",
         clinicaldf: pd.DataFrame,
         unmapped_oncotree_indices: pd.DataFrame,
     ) -> Tuple[str, str]:
-        """ This function returns the error and warning messages 
+        """This function returns the error and warning messages
         if the input clinical data has row indices with unmapped
         oncotree codes
 
--- a/genie_registry/clinical.py
+++ b/genie_registry/clinical.py
@@ -520,21 +520,17 @@ class Clinical(FileTypeFormat):
         errors = ""
         warnings = ""
         if len(unmapped_oncotree_indices) > 0:
+            # sort the unique unmapped oncotree codes
+            unmapped_oncotree_codes = sorted(
+                set(clinicaldf.loc[unmapped_oncotree_indices]["ONCOTREE_CODE"])
+            )
             errors = (
                 "Sample Clinical File: Please double check that all your "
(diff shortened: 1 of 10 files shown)


thomasyu888 (approved):
🔥 LGTM!  Nice - yes, admittedly - ordering in Python has always been an issue.  Thanks for doing this!

genie_registry/clinical.py
```diff
@@ -472,6 +472,68 @@ def process_steps(
...
+            )
+            errors = (
+                "Sample Clinical File: Please double check that all your "
+                "ONCOTREE CODES exist in the mapping. You have {} samples "
+                "that don't map. These are the codes that "
+                "don't map: {}\n".format(
+                    len(unmapped_oncotree_indices), ",".join(unmapped_oncotree_codes)
+                )
+            )
+        return errors, warnings
```
thomasyu888, line 535: Nit: at some point, I'd love for us to not return a tuple, but a namedtuple or something.  This has always bothered me since the beginning, but its a pretty significant code change.

rxu17: Created a ticket for this: https://sagebionetworks.jira.com/browse/GEN-1265 so it's on our radar

PR #560, opened 2024-04-22
[GEN-868] add year death validation

danlu1:
Problem:

Lack of validation check on whether YEAR_DEATH >= YEAR_CONTACT

Solution:

Add the validation check to compare YEAR_DEATH and YEAR_CONTACT when both of them exist and not NAs. 

Testing:
Unit test has been added.

### DIFF
--- a/genie/process_functions.py
+++ b/genie/process_functions.py
@@ -6,16 +6,15 @@ import logging
 import os
 import time
 from typing import Optional, Union
-import yaml
 
 import pandas as pd
 import requests
 import synapseclient
+import yaml
+from genie import extract
 from requests.adapters import HTTPAdapter
 from synapseclient import Synapse
 from urllib3.util import Retry
-
-from genie import extract
 
 pd.options.mode.chained_assignment = None
 
@@ -166,16 +165,16 @@ def checkUrl(url):
 
 
 # TODO Add to validate.py
-def checkColExist(DF: pd.DataFrame, key: Union[str, int]):
+def checkColExist(DF: pd.DataFrame, key: Union[str, int, list]) -> bool:
     """
-    This function checks if the column exists in a dataframe
+    This function checks if the column(s) exist(s) in a dataframe
 
     Args:
         DF: pandas dataframe
-        key: Expected column header name
+        key: Expected column header name(s)
 
     Returns:
-        bool:  True if column exists
+        bool:  True if column(s) exist(s)
     """
     result = False if DF.get(key) is None else True
     return result
--- a/genie_registry/clinical.py
+++ b/genie_registry/clinical.py
@@ -186,7 +186,7 @@ def _check_year_death_validity(clinicaldf: pd.DataFrame) -> str:
         Error message if YEAR_DEATH if invalid
     """
     error = ""
-    # generate temp dataframe to handle datatype mismatch in a column
+    # Generate temp dataframe to handle datatype mismatch in a column
     temp = clinicaldf.copy()
     # Convert YEAR_DEATH and YEAR_CONTACT to numeric, coercing errors to NaN
     temp["YEAR_DEATH"] = pd.to_numeric(temp["YEAR_DEATH"], errors="coerce")
--- a/genie_registry/clinical.py
+++ b/genie_registry/clinical.py
@@ -196,6 +196,7 @@ def _check_year_death_validity(clinicaldf: pd.DataFrame) -> str:
         "N/A",
         temp["YEAR_DEATH"] >= temp["YEAR_CONTACT"],
     )
+    index = [",".join(str(idx)) for idx, i in enumerate(check_result) if i == "False"]
     if "False" in check_result:
         error = "Patient Clinical File: Please double check your YEAR_DEATH and YEAR_CONTACT columns. YEAR_DEATH must be >= YEAR_CONTACT.\n"
     return error
--- a/genie_registry/clinical.py
+++ b/genie_registry/clinical.py
@@ -177,7 +177,11 @@ def _check_int_year_consistency(
 
 def _check_year_death_validity(clinicaldf: pd.DataFrame) -> str:
     """
-    Check if YEAR_DEATH >= YEAR_CONTACT
+    Check if YEAR_DEATH >= YEAR_CONTACT.
+    YEAR_DEATH should alway be equal or large than YEAR_CONTACT if they are both available.
+    This function compares rows with numeric values in both columns and returns comparion results("True"/"False")
+    If either of the column contains NA or nominal data (e.g. "Unknown", "Not Collected", "Unknown", "Not Applicable"),
+    then "N/A" will be outputed.
 
     Args:
         clinicaldf: Clinical Data Frame
@@ -196,8 +200,8 @@ def _check_year_death_validity(clinicaldf: pd.DataFrame) -> str:
         "N/A",
         temp["YEAR_DEATH"] >= temp["YEAR_CONTACT"],
     )
-    index = [",".join(str(idx)) for idx, i in enumerate(check_result) if i == "False"]
-    if "False" in check_result:
+    idx = [",".join(str(idx)) for idx, i in enumerate(check_result) if i == "False"]
+    if idx:
         error = "Patient Clinical File: Please double check your YEAR_DEATH and YEAR_CONTACT columns. YEAR_DEATH must be >= YEAR_CONTACT.\n"
     return error
 
@@ -851,10 +855,10 @@ class Clinical(FileTypeFormat):
         total_error.write(error)
 
         # CHECK: YEAR DEATH against YEAR CONTACT
-        haveColumn = process_functions.checkColExist(
+        has_death_and_contact_years = process_functions.checkColExist(
             clinicaldf, ["YEAR_DEATH", "YEAR_CONTACT"]
         )
-        if haveColumn:
+        if has_death_and_contact_years:
             error = _check_year_death_validity(clinicaldf=clinicaldf)
             total_error.write(error)
 
--- a/genie_registry/clinical.py
+++ b/genie_registry/clinical.py
@@ -5,7 +5,7 @@ import datetime
 import logging
 import os
 from io import StringIO
-from typing import Optional
+from typing import Optional, Tuple
 
 import numpy as np
 import pandas as pd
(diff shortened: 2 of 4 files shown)


genie_registry/clinical.py
```diff
@@ -174,7 +174,27 @@ def _check_int_year_consistency(
...
+    When YEAR_DEATH is known, YEAR_CONTACT should = YEAR_DEATH.
 
+    Args:
+        clinicaldf: Clinical Data Frame
+        
+    Returns:
+        Error message if YEAR_DEATH if invalid
+    """    
+    error = ""
+    try:
```
rxu17: I would recommend not using `try, except` here because we don't want to lose other exceptions that could occur other than the intentional `assert` statement.

genie/load.py
rxu17, line 1: I think you might have added in pandas 2.0 changes to this branch or created this branch from the pandas 2.0 branch? It should be created from `develop`.

genie_registry/clinical.py
```diff
@@ -823,6 +843,12 @@ def _validate(self, clinicaldf):
         )
         total_error.write(error)
 
+        # CHECK: YEAR DEATH against YEAR CONTACT
+        haveColumn = process_functions.checkColExist(clinicaldf, ["YEAR_DEATH", "YEAR_CONTACT"])
```
rxu17: I don't think `process.functions.checkColExist` accepts a list of keys. It would have to be an individual column/key you use.

genie_registry/clinical.py
```diff
@@ -174,7 +174,27 @@ def _check_int_year_consistency(
...
+    Returns:
+        Error message if YEAR_DEATH if invalid
+    """    
+    error = ""
+    try:
+        check_result = np.where((pd.isna(clinicaldf['YEAR_DEATH']) | pd.isna(clinicaldf['YEAR_CONTACT'])), 'N/A', clinicaldf['YEAR_DEATH'] >= clinicaldf['YEAR_CONTACT'])
+        assert not "False" in check_result
+    except Exception:
+        error = (
+            f"Please double check the YEAR_DEATH column. "
```
rxu17: Could you change the validation message to something like:
`"Patient Clinical File: Please double check your YEAR_DEATH and YEAR_CONTACT columns. YEAR_DEATH must be >= YEAR_CONTACT."`

rxu17 (commented):
Thanks for doing this! Just some initial comments.

Edit: Could you add more details to the PR description?

genie/load.py
danlu1: hmmm interesting, let me check. Thanks for catching.

genie_registry/clinical.py
danlu1: okay will tweak this.

danlu1:
> Thanks for doing this! Just some initial comments.
> 
> Edit: Could you add more details to the PR description?

Will do

genie_registry/clinical.py
danlu1: Sure.

genie_registry/clinical.py
danlu1: I tested it when I wrote this. The `checkColExist` output a dataframe with the filled column names and if the output dataframe is not empty it returns True otherwise False. This works when we want to check if both `YEAR_DEATH` and `YEAR_CONTACT` exist concurrently. Here are test cases I used:
Screenshot 2024-04-22 at 1 36 13 PM

rxu17: Neat! Could you update the docstring and type hints of `process_functions.checkColExist` and any tests (to have multiple column checks as test cases if not already) to reflect this? Thanks!

danlu1: Sure.

genie_registry/clinical.py
```diff
@@ -174,7 +174,27 @@ def _check_int_year_consistency(
 
     return ""
 
+def _check_year_death_validity(clinicaldf: pd.DataFrame) -> str:
+    """
+    Check if YEAR_DEATH >= YEAR_CONTACT.
+    When YEAR_DEATH is known, YEAR_CONTACT should = YEAR_DEATH.
```
rxu17: Per Chelsea's response to the JIRA ticket, can you update this docstring

genie_registry/clinical.py
```diff
@@ -174,7 +174,27 @@ def _check_int_year_consistency(
...
 
+    Args:
+        clinicaldf: Clinical Data Frame
+        
+    Returns:
+        Error message if YEAR_DEATH if invalid
+    """    
+    error = ""
+    try:
+        check_result = np.where((pd.isna(clinicaldf['YEAR_DEATH']) | pd.isna(clinicaldf['YEAR_CONTACT'])), 'N/A', clinicaldf['YEAR_DEATH'] >= clinicaldf['YEAR_CONTACT'])
```
rxu17: Sorry, this is not explicitly mentioned in the ticket but for all validation tickets, let's try to set up the validation for future row-based validation work. So having this give you the row indices (i've been returning pandas Index) that `YEAR_DEATH` < `YEAR_CONTACT`. And then your check will be on whether it's empty row indices or not

danlu1: Just added the line of code to show the index of the rows of YEAR_DEATH < YEAR_CONTACT for future use.

genie_registry/clinical.py
```diff
@@ -175,6 +175,33 @@ def _check_int_year_consistency(
...
+    temp["YEAR_DEATH"] = pd.to_numeric(temp["YEAR_DEATH"], errors="coerce")
+    temp["YEAR_CONTACT"] = pd.to_numeric(temp["YEAR_CONTACT"], errors="coerce")
+    check_result = np.where(
+        (pd.isna(temp["YEAR_DEATH"]) | pd.isna(temp["YEAR_CONTACT"])),
+        "N/A",
+        temp["YEAR_DEATH"] >= temp["YEAR_CONTACT"],
+    )
+    index = [",".join(str(idx)) for idx, i in enumerate(check_result) if i == "False"]
+    if "False" in check_result:
+        error = "Patient Clinical File: Please double check your YEAR_DEATH and YEAR_CONTACT columns. YEAR_DEATH must be >= YEAR_CONTACT.\n"
```
BryanFauble: Should you include the value(s) in the error message that caused the result to occur?

genie_registry/clinical.py
```diff
@@ -823,6 +850,14 @@ def _validate(self, clinicaldf):
         )
         total_error.write(error)
 
+        # CHECK: YEAR DEATH against YEAR CONTACT
+        haveColumn = process_functions.checkColExist(
```
BryanFauble: Nit: snake_case over camelCase

BryanFauble: Nit: I usually prefer current tense. Like has_column

genie_registry/clinical.py
```diff
@@ -823,6 +850,14 @@ def _validate(self, clinicaldf):
...
+            clinicaldf, ["YEAR_DEATH", "YEAR_CONTACT"]
+        )
+        if haveColumn:
+            error = _check_year_death_validity(clinicaldf=clinicaldf)
+            total_error.write(error)
+
```
BryanFauble, line 889: A stretch goal outside of the explicit scope of this change would be to split out these other items into their own functions. Is there a tech debt jira that already exists, or could one be created to address this item?

Cc: @rxu17

BryanFauble (approved):
Awesome work! Just a few formatting and follow up items I noted.

genie_registry/clinical.py
danlu1: From what I learned, we will eventually move forward to row-wise validation but this is the timeline is TBD.

genie_registry/clinical.py
danlu1: I like this idea. I use this variable name to match what we have in the existing code. @rxu17 we can convert to snake_case when we start refactoring the code.

thomasyu888: @danlu1, @rxu17 I am ok with iterative improvements to the code base as you're doing commits.  Its a balance, we don't want to go too out of the scope of the original PR, but certain things might be ok.  I'll leave it to your judgement how you want to tackle it, this could certainly be logged as a Jira ticket. 

Basically - see something fix something.  Every commit should be an improvement to the codebase.  Ofcourse there are things to think about like, are there other packages using this function.

genie_registry/clinical.py
danlu1: Sorry I missed the check row indices part. Will add it later.

genie_registry/clinical.py
danlu1: I'm happy to tweak as much as I can within the scope of the original PR. @rxu17 let me know your thoughts

genie_registry/clinical.py
rxu17: Yes we have a tech debt jira/part of the overarching goal for this. The goal is to eventually have a validation function for each validation rule for each fileformat especially when we want to implement GX framework for this.

I think this is part of our documentation epic as well. To break up the validation code so we can document each validation rule individually and completely.

This is not just happening here but throughout the genie code (also for processing as well) but we definitely want to start with validation first.

genie_registry/clinical.py
rxu17: @danlu1 Two comments:

- Let's use a more specific variable name and not reuse `haveColumn`, something like `has_contact_and_death_years` or related.
- Since this is a new variable you are creating that's not being used in the code elsewhere (other than your new code you're adding), feel free to use snake_case

Let me know what you think of that

genie_registry/clinical.py
rxu17: Thanks! I think you can still keep your check or check based on the row indices. It's enough to have the index of rows saved for later so thanks for adding that!

genie/process_functions.py
```diff
@@ -166,16 +165,16 @@ def checkUrl(url):
 
 
 # TODO Add to validate.py
-def checkColExist(DF: pd.DataFrame, key: Union[str, int]):
+def checkColExist(DF: pd.DataFrame, key: Union[str, int, list]) -> bool:
```
rxu17, line 168: I don't see any pre-existing tests for `checkColExist`, do you mind adding some to `tests/test_process_functions.py`? This is a bit outside the scope of this PR but the PR does depend on the expanded functionality of this function

genie_registry/clinical.py
```diff
@@ -175,6 +175,33 @@ def _check_int_year_consistency(
     return ""
 
 
+def _check_year_death_validity(clinicaldf: pd.DataFrame) -> str:
+    """
+    Check if YEAR_DEATH >= YEAR_CONTACT
```
rxu17: Could you add more to the description here? Details about what is considered invalid vs valid in this scenario.

genie_registry/clinical.py
rxu17: For more context: 

Reasoning why we wouldn't add the row-based validation (i think this is more helpful than just putting in values) in the error message yet just because we haven't decided to start implementing and delivering row-based validation for the centers/sites as a whole and we would have to talk about the standard for doing that so we are consistent across all validation rules. (And most likely discuss with the sites as a whole as well) so bigger discussions to be had there.

genie_registry/clinical.py
```diff
@@ -175,6 +175,33 @@ def _check_int_year_consistency(
...
+    temp = clinicaldf.copy()
+    # Convert YEAR_DEATH and YEAR_CONTACT to numeric, coercing errors to NaN
+    temp["YEAR_DEATH"] = pd.to_numeric(temp["YEAR_DEATH"], errors="coerce")
+    temp["YEAR_CONTACT"] = pd.to_numeric(temp["YEAR_CONTACT"], errors="coerce")
+    check_result = np.where(
+        (pd.isna(temp["YEAR_DEATH"]) | pd.isna(temp["YEAR_CONTACT"])),
+        "N/A",
+        temp["YEAR_DEATH"] >= temp["YEAR_CONTACT"],
+    )
+    index = [",".join(str(idx)) for idx, i in enumerate(check_result) if i == "False"]
```
rxu17: Wondering why you're joining the indices via "," as a string? I think maybe you meant to do something like this:
`indices = [idx for idx, result in enumerate(check_result) if result is False]`

What I've found helpful in the past is breaking up the validation functions into two parts: one function for outputting the invalid row indices and one function for outputting the error message so you can run tests on the actual invalid results.

genie_registry/clinical.py
```diff
@@ -175,6 +175,37 @@ def _check_int_year_consistency(
...
+
+    Args:
+        clinicaldf: Clinical Data Frame
+
+    Returns:
+        Error message if YEAR_DEATH if invalid
+    """
+    error = ""
+    # Generate temp dataframe to handle datatype mismatch in a column
+    temp = clinicaldf.copy()
```
thomasyu888: Nit: to reduce the amount of memory used, you may want to scope this to copy only the two columns in question.

You're also going to have another task with `INT_DOD` >= `INT_CONTACT`. No need to change anything here, but you may want to when you tackle the other function so you don't duplicate code.  Minor detail though.

thomasyu888 (approved):
🔥 I'm going to pre-approve!  great work here!

genie_registry/clinical.py
```diff
@@ -175,6 +175,37 @@ def _check_int_year_consistency(
     return ""
 
 
+def _check_year_death_validity(clinicaldf: pd.DataFrame) -> str:
+    """
+    Check if YEAR_DEATH >= YEAR_CONTACT.
+    YEAR_DEATH should alway be equal or large than YEAR_CONTACT if they are both available.
+    This function compares rows with numeric values in both columns and returns comparion results("True"/"False")
```
rxu17: Nit: I think this statement here is a bit confusing because it says that the function itself is returning true/false/'N/A' values? 

I think the statement is helpful to be added in comments down in your function below above the relevant sections of code because it's not obvious what the code is doing, and the statement clears that up

rxu17 (approved):
Also going to pre-approve! (just have the row-based validation implementation and addressing some of the comments)

PR #561, opened 2024-04-25
[GEN-867] Add validation rule to check if INT_DOD >= INT_CONTACT

danlu1:
Problem:

Lack of validation check on whether INT_DOD >= INT_CONTACT

Solution:

Add the validation check to compare INT_DOD and INT_CONTACT when both of them exist and not NAs.

Testing:
Unit test has been added.

### DIFF
--- a/genie_registry/clinical.py
+++ b/genie_registry/clinical.py
@@ -191,9 +191,7 @@ def _check_year_death_validity(clinicaldf: pd.DataFrame) -> pd.Index:
     # Convert YEAR_DEATH and YEAR_CONTACT to numeric, coercing errors to NaN
     temp["YEAR_DEATH"] = pd.to_numeric(temp["YEAR_DEATH"], errors="coerce")
     temp["YEAR_CONTACT"] = pd.to_numeric(temp["YEAR_CONTACT"], errors="coerce")
-    # Compare rows with numeric values in both columns and returns comparion results("True"/"False")
-    # If either of the column contains NA or nominal data (e.g. "Unknown", "Not Collected", "Not Applicable"),
-    # "N/A" will be outputed.
+    # Compare rows with numeric values in both YEAR_DEATH and YEAR_CONTACT columns
     temp["check_result"] = np.where(
         (pd.isna(temp["YEAR_DEATH"]) | pd.isna(temp["YEAR_CONTACT"])),
         "N/A",
@@ -223,7 +221,7 @@ def _check_year_death_validity_message(
             "Patient Clinical File: Please double check your YEAR_DEATH and YEAR_CONTACT columns. "
             "YEAR_DEATH must be >= YEAR_CONTACT. "
             f"There are {len(invalid_year_death_indices)} row(s) with YEAR_DEATH < YEAR_CONTACT. "
-            f"Row {invalid_year_death_indices.tolist()} contain invalid values in the YEAR_DEATH field. Please correct.\n"
+            f"The row number(s) this occurs in are: {invalid_year_death_indices.tolist()}. Please correct.\n"
         )
     return error, warning
 
@@ -244,9 +242,7 @@ def _check_int_dod_validity(clinicaldf: pd.DataFrame) -> pd.Index:
     # Convert INT_DOD and INT_CONTACT to numeric, coercing errors to NaN
     temp["INT_DOD"] = pd.to_numeric(temp["INT_DOD"], errors="coerce")
     temp["INT_CONTACT"] = pd.to_numeric(temp["INT_CONTACT"], errors="coerce")
-    # Compare rows with numeric values in both columns and returns comparion results("True"/"False")
-    # If either of the column contains NA or nominal data (e.g. "Unknown", "Not Collected", "Not Applicable"),
-    # "N/A" will be outputed.
+    # Compare rows with numeric values in both INT_DOD and INT_CONTACT columns
     temp["check_result"] = np.where(
         (pd.isna(temp["INT_DOD"]) | pd.isna(temp["INT_CONTACT"])),
         "N/A",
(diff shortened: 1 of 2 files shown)


genie_registry/clinical.py
```diff
@@ -228,6 +228,59 @@ def _check_year_death_validity_message(
...
+    INT_DOD should alway be greater than or equal to INT_CONTACT when they are both available.
+    This function checks if INT_DOD >= INT_CONTACT and returns row indices of invalid INT_DOD rows.
+
+    Args:
+        clinicaldf: Clinical Data Frame
+
+    Returns:
+        pd.Index: The row indices of the row with INT_DOD < INT_CONTACT in the input clinical data
+    """
+    # Generate temp dataframe to handle datatype mismatch in a column
```
BryanFauble, line 240: Try to limit the amount of in-line comments you add. In a lot of cases it has the unintended affect of reducing readability. The code you wrote does well at telling us what the code is doing. That should be the source of truth.

genie_registry/clinical.py
```diff
@@ -228,6 +228,59 @@ def _check_year_death_validity_message(
...
+    temp = clinicaldf[["INT_DOD", "INT_CONTACT"]].copy()
+    # Convert INT_DOD and INT_CONTACT to numeric, coercing errors to NaN
+    temp["INT_DOD"] = pd.to_numeric(temp["INT_DOD"], errors="coerce")
+    temp["INT_CONTACT"] = pd.to_numeric(temp["INT_CONTACT"], errors="coerce")
+    # Compare rows with numeric values in both columns and returns comparion results("True"/"False")
```
BryanFauble: This inline comment is misleading. You are only checking for False.

genie_registry/clinical.py
```diff
@@ -228,6 +228,59 @@ def _check_year_death_validity_message(
...
+        (pd.isna(temp["INT_DOD"]) | pd.isna(temp["INT_CONTACT"])),
+        "N/A",
+        temp["INT_DOD"] >= temp["INT_CONTACT"],
+    )
+    invalid_int_dod = temp[temp["check_result"] == "False"]
+    return invalid_int_dod.index
+
+
+def _check_int_dod_validity_message(
+    invalid_int_dod_indices: pd.Index,
```
BryanFauble, line 256: Is the type here a list of pd.Index? The length check below suggests it's an iterator of values.

BryanFauble: Might also just be the pandas specific way of doing things that is not super intuitive

genie_registry/clinical.py
```diff
@@ -956,6 +1009,16 @@ def _validate(self, clinicaldf):
...
         else:
             total_error.write("Patient Clinical File: Must have DEAD column.\n")
+
+        # CHECK: INT DOD against INT CONTACT
+        has_int_dod_and_contact = process_functions.checkColExist(
+            clinicaldf, ["INT_DOD", "INT_CONTACT"]
+        )
+        if has_int_dod_and_contact:
+            invalid_int_dod_indices = _check_int_dod_validity(clinicaldf)
+            errors, warnings = _check_int_dod_validity_message(invalid_int_dod_indices)
```
BryanFauble, line 1015: Nit:
`errors, _`
As warnings is not used

BryanFauble (approved):
A few formatting suggestions, otherwise the business logic looks good!

genie_registry/clinical.py
```diff
@@ -228,6 +228,59 @@ def _check_year_death_validity_message(
...
+        input clinical data has
+    """
+    error = ""
+    warning = ""
+    if len(invalid_int_dod_indices) > 0:
+        error = (
+            "Patient Clinical File: Please double check your INT_DOD and INT_CONTACT columns. "
+            "INT_DOD must be >= INT_CONTACT. "
+            f"There are {len(invalid_int_dod_indices)} row(s) with INT_DOD < INT_CONTACT. "
+            f"Row {invalid_int_dod_indices.tolist()} contain invalid values in the INT_DOD field. Please correct.\n"
```
rxu17: This part of the message is bit misleading because it implies that the `INT_DOD` field is the issue when it could be `INT_CONTACT`. You could rephrase this part of the error message to be like:
`The row number(s) this occurs in are: {invalid_year_death_indices.tolist()}. Please correct.`

I think this is happening for the `YEAR_DEATH` AND `YEAR_CONTACT` validation rule function too.

rxu17 (approved):
LGTM! Just one comment

genie_registry/clinical.py
thomasyu888: ^ Please fix the above in this PR as well.  Thanks!

thomasyu888 (approved):
🔥 Fantastic work, and great code reviews!  Thanks team - will pre-approve.

genie_registry/clinical.py
danlu1: Sure. Will do.

genie_registry/clinical.py
danlu1: Hmmm let me rephrase this.

genie_registry/clinical.py
danlu1: The type is `pandas.core.indexes.range.RangeIndex`. We can either use `len` or `size`. 
Screenshot 2024-04-26 at 11 31 37 AM

genie_registry/clinical.py
danlu1: This is adapted from https://github.com/Sage-Bionetworks/Genie/blob/3f318022e1ef3e5de8c774ce79f7b521ab5653b4/genie_registry/clinical.py#L475-L535 and we might need to output warnings in the future.

PR #562, opened 2024-05-01
[GEN-1263] Fix bed file duplicates

rxu17:
**Purpose:** This is a draft PR. This PR fixes the bug that always appends each bed file to the bed synapse database table during processing creating duplicates in the bed database table.

**Changes:**
- [Main fix is in bed.py in `process_steps`](https://github.com/Sage-Bionetworks/Genie/blob/gen-1263-fix-bed-file-dups/genie_registry/bed.py#L609-L610)

**Tests:**

- Added unit tests (also added tests to the `update_table` function as well since there wasn't any present
- Ran bed file through processing and no bed file rows were appended to the test bed database table here: https://www.synapse.org/#!Synapse:syn11600834/
- Ran bed file on NF tower using [updated docker image here](https://github.com/Sage-Bionetworks/Genie/pkgs/container/genie/213829252?tag=gen-1285-remove-row-based-validation) twice, and ran [pipeline comparison](https://github.com/Sage-Bionetworks/Genie_processing?tab=readme-ov-file#pipeline-version-comparisons) on it - no changes. [Bed database table here](https://www.synapse.org/#!Synapse:syn11600834/tables/)

Depends on #565

### DIFF
--- a/genie/load.py
+++ b/genie/load.py
@@ -162,7 +162,7 @@ def update_table(
         syn=syn,
         database=database,
         new_dataset=newData,
-        database_syn_id=databaseSynId,
+        database_synid=databaseSynId,
         primary_key_cols=databaseEnt.primaryKey,
         to_delete=toDelete,
     )
--- a/genie/load.py
+++ b/genie/load.py
@@ -159,7 +159,12 @@ def update_table(
     else:
         newData = newData[database.columns]
     _update_table(
-        syn, database, newData, databaseSynId, databaseEnt.primaryKey, toDelete
+        syn=syn,
+        database=database,
+        new_dataset=newData,
+        database_syn_id=databaseSynId,
+        primary_key_cols=databaseEnt.primaryKey,
+        to_delete=toDelete,
     )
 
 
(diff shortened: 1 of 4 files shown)


thomasyu888 (approved):
🔥 LGTM!  Thanks for adding tests and for the fix!

Although, not sure why the docker build is failing...

rxu17:
> Although, not sure why the docker build is failing...

Yeah, seems to do with compilation issues during the installation of `rlang` and them enforcing warnings as errors (funny, no issue a couple days ago)

rxu17:
@thomasyu888 Should this be directly merged to `main` because we have our upcoming release? I forgot our decision on this.

thomasyu888:
@rxu17 hmmm... it might be best to push to develop and wait for all the changes for this release. One reason is because the docker build fix is in develop, so pushing directly to main may break the docker build.

Thoughts?

We should probably create a ticket to manually drop dups from the bed table when we are ready to push everything up.  There's a bit of risk of releasing before the public release, so we can discuss whether we should or not.

rxu17:
Thanks for the reminder about the docker build (I realize we didn't push that directly to main as well which makes sense as well). Will just merge to `develop` for now

PR #563, opened 2024-05-01
[GEN-1285] Remove row-based validation error message

danlu1:
Problem:
The row-based error message is somewhat misleading since the row indices are from the merged data frame of patient and sample data instead of the patient clinical file or sample clinical file. 

Solution:
Remove row-based validation

Test:
Affected unit tests are updated.


thomasyu888 (approved):
🔥 LGTM!

PR #564, opened 2024-05-01
[GEN-1285] Remove row based validation

danlu1:
Problem:

No line breaks after INT_DOD and YEAR_DEATH validity checks.

Solution:
Added line breaks.

Testing:
Updated affected unit tests


thomasyu888 (approved):
🔥 LGTM! Forgot about the line breaks.....

Let fix the linting rules, and also make sure the container builds :)

PR #565, opened 2024-05-04
[GEN-1294] pin r-base version and upgrade to Node 20

danlu1:
Problem:

The Build_Docker step in the ci pipeline failed recently and ran into the Error: install of package 'rlang' failed [error code 1]. I figured the root cause for the pipeline breaks because that the [newest r-base (4.4.0) ](https://www.r-bloggers.com/2024/04/whats-new-in-r-4-4-0/) was released 4/25/2024. The apt-install pulled this version of r-base which is incompatible with the gcc-9.

Solution:

Pin r-base version to 4.3 and update actions of Node 16 with Node 20.

Testing:

Tested the workflow locally and it run through successfully.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,9 +25,9 @@ jobs:
       matrix:
         python-version: [3.8, 3.9]
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v4
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v4
+      uses: actions/setup-python@v5
       with:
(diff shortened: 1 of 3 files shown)


thomasyu888 (approved):
🔥 LGTM, this keeps the python at 3.8 right?

rxu17 (approved):
LGTM! Just one question. Did you run the entire test pipeline successfully on the nextflow tower using this branch's docker image?

danlu1:
@thomasyu888 I pined Python3.8. @rxu17 yes, this is fully tested on both EC2 and Tower.

PR #567, opened 2024-05-17
[GEN-1313] Export detailed columns for NAACCR codes

danlu1:
Problem: 

Need to export detailed columns for NAACCR codes

Solution:

Add the detailed columns when corresponding race, sex, and ethnicity columns are available.

Test:

Tested locally and ran through validation and processing steps successfully with expected results.

### DIFF
--- a/genie/consortium_to_public.py
+++ b/genie/consortium_to_public.py
@@ -139,10 +139,11 @@ def consortiumToPublic(
     )
 
     # Clinical release scope filter
-    # If consortium -> Don't release to public
-    # TODO: check why this synapse id is hard coded?
+    clinical_tier_release_scope_synid = databaseSynIdMappingDf["Id"][
+        databaseSynIdMappingDf["Database"] == "clinical_tier_release_scope"
+    ][0]
     publicRelease = extract.get_syntabledf(
-        syn=syn, query_string="SELECT * FROM syn8545211 where releaseScope = 'public'"
+        syn=syn, query_string=f"SELECT * FROM {clinical_tier_release_scope_synid} where releaseScope = 'public'"
     )
 
     allClin = clinicalDf[clinicalDf["SAMPLE_ID"].isin(publicReleaseSamples)]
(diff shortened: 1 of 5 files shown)


rxu17 (commented):
Code looks good so far! i would just adjust the tests so they pass

danlu1:
Close pr for now to rebuild docker image

tests/test_clinical.py
```diff
@@ -956,8 +956,15 @@ def test_remap_clinical_values_sampletype():
 )
 def test_remap_clinical_values(col):
     """Test Remapping clinical values"""
-    testdf = pd.DataFrame({col: [1, 2, 99]})
-    expecteddf = pd.DataFrame({col: ["Male", "Female", "Unknown"]})
+    testdf = pd.DataFrame({"SEX": [1, 2, 99], "PRIMARY_RACE": [1, 2, 99]})
```
rxu17: I would add a couple more tests here:

- test for having all code columns (to make the detailed cols) you could have so we make sure it remaps for all of them
- test for when you have none of the columns

tests/test_clinical.py
rxu17, line 1: Could you add in a test for the clinical file's `preprocess` function since you've already written the test for it previously? You previously did this for [this PR](https://github.com/Sage-Bionetworks/Genie/pull/554/files#diff-4890ac1183096b3845539abb5712c0ed0b70a0ed8ddb7a2768011aa87981aa09) which is part of pandas 2.0 updates so you could reuse most of the code from this.

genie/consortium_to_public.py
rxu17, line 1: Just a comment. Nothing actionable here. I think it's OK we don't have unit tests for these giant functions: `stagingToCbio`, `store_clinical_files` and `consortiumToPublic` in the `database_to_staging` and `consortium_to_public` code just because these are better tested using integration tests and we already did that with our pipeline comparisons and our test runs on the test pipeline. 

Doing this would be outside the scope of this ticket and is already part of our tech debt epic to refactor and add tests.

rxu17 (approved):
LGTM! Great work! Just a couple of comments but going to pre-approve. Also just a heads up, i think you have to merge in changes from `develop`

genie/consortium_to_public.py
danlu1: Yes, I agree that we add unit tests later for these functions for our tech debt epic.

thomasyu888: The more we can spin out "unit functions" within these larger functions, the more they will be better served as integration tests. 

If we think of these as ETL, it should tell the story of data processing.

tests/test_clinical.py
danlu1: Done.

PR #568, opened 2024-06-01
[GEN-1330] Setup staging pipeline

rxu17:
**Purpose:** This is a draft PR. We are trying to set up the staging part of the pipeline so we can run production data in the staging project and test out changes to our pipeline without running it on the production project.

There were some relatively big refactors needed in the maf in bed and mutation_in_cis filter functions in order to have it run on the staging project vs on the production project (which it was doing before running on production project despite having the `--staging` flag set)

**Changes** Main refactor changes (breaking it down into individual functions) were made in `database_to_staging.py` in the (doesn't affect the functionality of the code): 

- [runMAFinBED.py](https://github.com/Sage-Bionetworks/Genie/blob/708f7ff2e3732c5517c87658a9ddcbb4d68d6c54/genie/database_to_staging.py#L414)
- [mutation_in_cis_filter](https://github.com/Sage-Bionetworks/Genie/blob/708f7ff2e3732c5517c87658a9ddcbb4d68d6c54/genie/database_to_staging.py#L577)

**Testing:** 

- [X] - Tests were added for all affected functions (nonexistent previously)
- [x] - Ran the `python bin/database_to_staging.py Jan-2017 ../cbioportal TEST --test` command and ran the pipeline comparison script `run_pipeline_version_comparisons.py` in order to make sure there were no changes to the relevant maf in bed and mutation in cis files when running in test mode - No changes were found
- [x] - Ran the consortium release step only option on `TEST` pipeline in NF ([See code here that supports that](https://github.com/Sage-Bionetworks-Workflows/nf-genie/pull/19)) using docker image rebuilt from this feature branch against `develop` image, ran comparison script - no differences
- [x] - Ran the public release step only option on `TEST` pipeline in NF ([See code here that supports that](https://github.com/Sage-Bionetworks-Workflows/nf-genie/pull/19)) using docker image rebuilt from this feature branch against `develop` image, ran comparison script - no differences
- [x] - This more of a pre-req for the next two steps. Run the processing step for both mutation and main data on `STAGING` pipeline in NF.
- [x] - Ran the consortium release step only option on `STAGING` pipeline in NF ([See code here that supports that](https://github.com/Sage-Bionetworks-Workflows/nf-genie/pull/19)) using docker image rebuilt from this feature branch against `develop` image, ran comparison script - no unexpected differences

Decided we will not test public release. Ran the public release step only option on `STAGING` pipeline in NF ([See code here that supports that](https://github.com/Sage-Bionetworks-Workflows/nf-genie/pull/19)) using docker image rebuilt from this feature branch against `develop` image


thomasyu888 (commented):
🔥 LGTM! @danlu1 , can you give this a look?

genie/database_to_staging.py
```diff
@@ -529,48 +601,88 @@ def mutation_in_cis_filter(
...
     """
-    if not skipMutationsInCis:
-        mergeCheck_script = os.path.join(
-            os.path.dirname(os.path.abspath(__file__)), "../R/mergeCheck.R"
-        )
-        command = ["Rscript", mergeCheck_script]
-        if test:
-            command.append("--testing")
+    if (not skipMutationsInCis) and (not staging):
+        command = get_mutation_in_cis_filter_script_cmd(test=test)
```
danlu1, line 610: Are you going to add a staging parameter for this function later when the staging mode is ready?

tests/test_database_to_staging.py
```diff
@@ -106,3 +109,735 @@ def test_store_assay_info_files(syn):
...
+            "name": "testing_true",
+            "genie_version": "TEST",
+            "variant_filtering_synId": "synZZZZZ",
+            "clinicaldf": pd.DataFrame(dict(SAMPLE_ID=["SAGE-1"])),
+            "beddf": pd.DataFrame(dict(SAMPLE_ID=["TEST-1"])),
+            "center_mappingdf": pd.DataFrame(dict(vcf2maf=["synZZZZZ"])),
+            "processing_date": "TEST_DATE",
+            "skip_mutationsincis": True,
+            "consortium_release_cutoff": "CUTOFF_DATE",
+            "test": False,
```
danlu1: Should this be `"test": True` and `"current_release_staging": False` so the `MAFinBED.R` will use staging databaseSynIdMappingId?

danlu1 (changes requested):
ahhh forgot to submit the review.

genie/database_to_staging.py
rxu17: Nope originally in the code we didn't run mutation in cis as part of staging so I left it as that as well because we don't need to implement that since it doesn't affect the clinical columns.

tests/test_database_to_staging.py
rxu17: Good catch, this test case should have testing : True to test the condition that MAFinBED.R will use the **test** `databaseSynIdMappingId`. 

The test case above this would test that MAFinBED.R will use the **staging** `databaseSynIdMappingId`.

danlu1 (approved):
LGTM.

thomasyu888 (approved):
🔥 LGTM!  I'm going to pre-approve, but let me know if you run into issues with doing the last 2 checks.

rxu17:
> but let me know if you run into issues with doing the last 2 checks.

Yup I'll probably be running all the checks minus the 1st again because i'm adding in the new mapping table changes + adding the check to not run the error email sending out for staging project.

templates/dashboardTemplate.Rmd
```diff
@@ -213,23 +213,30 @@ get_release_synid <- function(release_folder_fileview_synid, release) {
...
-  } else if (nrow_df == 1) {
-    public_reldf$id
-  } else {
-    # The releases are ordered by latest first, so
-    # always take the first element
-    public_reldf$id[1]
+  # HACK: until we create tables for these hardcoded synids in
+  # TEST and STAGING projects, will have to return staging 
+  # version of the public release folder here
+  if (grepl("STAGING", release)){
```
thomasyu888, line 205: Hmm I thought we had a test public release folder, does this mean our testing pipeline now writes to this?

thomasyu888: Edit: nevermind, we take this code to calculate two things
1. retraction
2. number of samples added

So the dashboard won't be fully reliable to compare since couple of things

1. We don't run the staging pipeline every consortium release
2.  The test pipeline doesn't have the concept of different consortium releases.

tests/test_database_to_staging.py
thomasyu888, line 1: Thanks for adding these tests!

templates/dashboardTemplate.Rmd
rxu17: We do have a test public release folder but we don't have testing for dashboard in our test project. So this doesn't run for testing, [see here](https://github.com/Sage-Bionetworks/Genie/blob/9295c5c2aec47a8a62282e26eac56736a4925edb/bin/database_to_staging.py#L289-L296)

R/mergeCheck.R
thomasyu888, line 1: Unfortunate that we have to add the --staging flag this way, but when we are ready to shift into using the project_id, we will at least have comparison code to run.

genie/database_to_staging.py
```diff
@@ -442,6 +447,64 @@ def runMAFinBED(syn, center_mappingdf, test=False, genieVersion="test"):
...
+    syn: synapseclient.Synapse,
+    notinbed_file: str,
+    center_mapping_df: pd.DataFrame,
+    genie_version: str,
+) -> pd.DataFrame:
+    """This script retrieves and stores the filtered variants generated
+        from running the MAFinBed script as files on Synapse. This script
+        then returns the not in bed file with the removed variants column
+        added.
+        TODO: Add handling for empty file?
```
thomasyu888, line 496: Good catch!  Is the notinbed_file sometimes empty?

genie/database_to_staging.py
```diff
@@ -529,48 +601,88 @@ def mutation_in_cis_filter(
...
+
+
+def get_mutation_in_cis_flagged_variants(
+    syn: synapseclient.Synapse, variant_filtering_synId: str
+) -> pd.Series:
+    """Gets the flagged variants in our variant filtering table which is
+       a unique string concatenation of the Chromosome, Start_Position,
+       HGVSp_Short, Reference_Allele, Tumor_Seq_Allele2 and Tumor_Sample_Barcode
+       columns
+       TODO: Add handling for when we have 0 row query results
```
thomasyu888, line 677: These are great TODOs to indicate weaknesses in the pipeline, but in the grand scheme of things, it is probably extremely rare moving forward that any of these tables are blank.

thomasyu888 (approved):
🔥 LGTM!  One thing to note for writing tests in general, which you may want to consult Bryan on.

When I wrote the GENIE tests, I was a bit of an amateur and a lot of the ones I wrote are too tied to the implementation.  What I mean by that is if you change some of the code now, the tests may also fail even if you haven't changed the behavior (which would require changes to the tests).  Good tests should test the behavior (sorry about  my less than ideal examples!)

A lot of this has made me realized that testing for data pipelines probably should be done:

1. Unit tests for the unit functions
2. integration tests will compose of ability to run the pipeline, then making sure data in vs data out and others.

Anyhow, fantastic work, and I'm going to pre-approve!

genie/database_to_staging.py
rxu17: I don't think so, that would mean we have 0 variants that don't overlap with any of the bed regions ... ? How likely that is genomically I wouldn't know but if the code is broken, something like that could happen

genie/database_to_staging.py
rxu17: Yeah this honestly should be a data quality check / expectation instead since it's directly data related. That this query should always return results.

genie/database_to_staging.py
thomasyu888: Very unlikely but good to think about

rxu17:
> When I wrote the GENIE tests, I was a bit of an amateur and a lot of the ones I wrote are too tied to the implementation. What I mean by that is if you change some of the code now, the tests may also fail even if you haven't changed the behavior (which would require changes to the tests). Good tests should test the behavior (sorry about my less than ideal examples!)

Good point. When I was initially writing this, I was worried about the refactoring of the code so I wanted to check every call in the code and at the time I wasn't familiar with the behavior of this code so we ended up with more tests tied to the direct implementation. Looking at it now, it makes more sense to test the end behavior / refactor it so there is an end behavior because most of these behaviors would be tested via integration test (like storing of files). I think we can definitely do this easier when we convert mafinbed to python, making it easier to just test / rebuild from there.

PR #572, opened 2024-07-02
[GEN-1348] Allow in mapping tables

rxu17:
**Purpose:** We want to add test versions of the ETHNICITY, SAMPLE_TYPE, RACE and SEX mapping tables so that whenever we add updates to these tables, we're testing the updates out first before modifying the production versions.

**Testing:**
- Run test pipeline with docker image with changes, and check that the new updates made it into the descriptions
- Run the pipeline comparison script between the pipeline run on this PR's docker image and the `develop` docker image
We see expected differences in the clinical patient files at the `staging` and `consortium_release` steps. We only see the changes in the `data_clinical.txt` file at the `public_release` step but not the `data_clinical_patient.txt`. This is expected

### DIFF
--- a/genie/extract.py
+++ b/genie/extract.py
@@ -279,12 +279,6 @@ def get_genie_config(
     center_mapping_df.index = center_mapping_df.center
     # Add center configurations including input/staging synapse ids
     genie_config["center_config"] = center_mapping_df.to_dict("index")
-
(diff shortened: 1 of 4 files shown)


thomasyu888 (approved):
🔥 LGTM!

PR #574, opened 2024-07-09
[GEN-703] Validation status with version

thomasyu888:
Problem:
Version number wasn't included in the validation status table

Solution:
Add version number. This will be a multi-process PR. We will need to either add new validation status tables, OR edit the schemas.

Extras:
Only download files that need to be validated.  This should speed up the download times. 

Future work:
Use fileview to get a list of files


genie/input_to_database.py
```diff
@@ -336,8 +335,13 @@ def validatefile(
         ancillary_files=ancillary_files,
     )
     filetype = validator.file_type
-
     if check_file_status["to_validate"]:
+        # HACK: Don't download again if only_validate is not True, but all
```
rxu17, line 339: 😭😭😭

tests/test_input_to_database.py
rxu17, line 1: Should we add to the tests (the ones that call `validatefile`) for the hack added for not downloading files again when we do validation + processing?

genie/input_to_database.py
```diff
@@ -868,8 +884,13 @@ def center_input_to_database(
...
     center_input_synid = genie_config["center_config"][center]["inputSynId"]
     logger.info("Center: " + center)
+    # Don't download all the files when only validate
     center_files = extract.get_center_input_files(
-        syn, center_input_synid, center, process
+        syn=syn,
+        synid=center_input_synid,
+        center=center,
+        process=process,
+        downloadFile=(not only_validate),
```
rxu17: hmm don't we need the files to download if we are only validating? Each validation run will have different temp. dir in the synapse cache and we want to make sure we're pulling the latest version of the file?

I get that 
```
if entities[0].get("path") is None:
            validator.entitylist = [syn.get(entity) for entity in entities]
```
is meant to just handle the situations where we want to just download the file once (e.g: when we run validation + processing in one go). Couldn't it just be`downloadFile = False` because the snippet of code will handle all situations? validation only, processing (where we run validation beforehand and download all the file)

rxu17 (approved):
Thanks for doing this! LGTM! I had a few comments/questions

genie/input_to_database.py
thomasyu888: 😅😅😅

genie/input_to_database.py
thomasyu888: So we actually don't want to download all the files when we validate the files, because there is validation logic to check if we need to actually validate the files.

1. If the entity name is changed
2. If the file md5 is changed

Against what is on the validation status table.

tests/test_input_to_database.py
thomasyu888: We should, but.... I'm hoping we can remove that hack after we:

1. Split validation + processing
2. Compare fileview  vs validation status table to get download all the files required for validation.

genie/input_to_database.py
thomasyu888: One thing to note here is that if someone only uploads the sample file, it's not going to download the maf or bed files so cross file validation might fail without those files downloaded.

That does raise the complexity of the code a bit....

So... We may want to comment out the hack first to avoid unwanted issues

rxu17: i agree, let's comment it out first and handle it in another ticket

thomasyu888:
Trying to figure out what's wrong with the linting... (nevermind, you got it @rxu17 !)

PR #575, opened 2024-07-11
Update docstring for cna's validateSymbol function

rxu17:
**Purpose:** Inspired by [GEN-1416](https://sagebionetworks.jira.com/browse/GEN-1416) and hopefully to reduce further confusion on this cna gene validation in the future.

[GEN-1416]: https://sagebionetworks.jira.com/browse/GEN-1416?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/genie_registry/cna.py
+++ b/genie_registry/cna.py
@@ -27,14 +27,15 @@ def validateSymbol(
         table's Hugo_Symbol value for the purpose of validation. The ID column is the
         original Hugo_Symbol column of the bed files before the Hugo_Symbol column gets
         mapped to valid possible gene values in the Actual Gene Positions (GRCh37)
-        database table.
+        database table. See the bed fileformat module's remap_symbols function and
+        how it gets used in processing for more info on this.
 
     The validation throws a WARNING if the gene doesn't satisfy
     either of the above two conditions
 
     Args:
         gene: Gene name
-        bedDf: the bed database table as pandas dataframe
+        bedDf: The bed database table as a pandas dataframe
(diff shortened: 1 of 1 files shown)


genie_registry/cna.py
```diff
@@ -10,17 +11,37 @@
...
     Args:
         gene: Gene name
-        bedDf: Bed pandas dataframe
+        bedDf: The bed database table as a pandas dataframe
         returnMappedDf: Return a mapped gene. Defaults to True
 
     Returns:
-        gene name or boolean for whether a gene is valid
+        Union[str, float, bool]:
+        Returns gene symbol (str if valid, a float("nan") if invalid) if returnMappedDf is True
```
BryanFauble, line 43: More of a comment on the complexity of the return type. Your changes are great to expand on this!

I would think we might want to have bool return from an `isValid` function. Then if it's valid a subsequent function to return the actual gene. Unfortunately it would be a lot larger of a change, but this has a bit of complexity where the caller has to check if it's a bool, then the value of that bool, and check if it's a float, and check if it's a string.

BryanFauble (approved):
Great change to add some clarity here!

genie_registry/cna.py
rxu17: Agreed, when I first looked at this, it was difficult to comprehend due to the return types! This is on the docket to be revised, hopefully we can get to it sooner

PR #576, opened 2024-07-15
[GEN-1429] Add variant_classification as known string col

rxu17:
**Purpose:** See [Jira ticket](https://sagebionetworks.jira.com/browse/GEN-1429) for more background info. An input maf file had `Variant_Classification` as a column (it's optional) and the first 5000 rows of it was NA so our processing maf code didn't deduce that it was a string column.

**Testing:**
- [X] Added unit tests for the functions using the `KNOWN_STRING_COLS`
- [X] Added `Variant_Classification` col with all NAs to test project
- [X] Ran on test pipeline locally
- [x] Ran on test pipeline via Tower

### DIFF
--- a/genie/process_mutation.py
+++ b/genie/process_mutation.py
@@ -115,6 +115,7 @@ KNOWN_STRING_COLS = [
     "Exon_Number",
     "genomic_location_explanation",
     "Annotation_Status",
+    "Variant_Classification",
 ]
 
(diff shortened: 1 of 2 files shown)


thomasyu888 (approved):
🔥 LGTM! NA's and pandas inferring dtype strikes again...

PR #578, opened 2024-10-01
Update configure maf docstring

rxu17:
**Purpose:** Small PR to add more details to the maf function that does germline variant filtering during consortium release step.


danlu1 (approved):
Thanks for adding this. It's very helpful.

PR #579, opened 2024-10-29
[GEN-1545] Remove GERMLINE from valid values of the SV_STATUS

danlu1:
Problem:

We have no need to collect germline structural variants and we didn't inform the data uploader about this when validating the data.

Solution:

Update the validation check for SV_STATUS and remove GERMLINE from valid values so when sites submit a germline variant, we flag that variant to be removed.

Testing:

Unit test has been added and integrated test has been done on EC2 and Tower.

Screenshot 2024-10-29 at 10 39 32 AM

### DIFF
--- a/genie_registry/structural_variant.py
+++ b/genie_registry/structural_variant.py
@@ -1,11 +1,10 @@
-from io import StringIO
 import logging
 import os
+from io import StringIO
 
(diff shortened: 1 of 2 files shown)


rxu17 (approved):
LGTM!

PR #581, opened 2024-11-26
[GEN-1622] remove sample class filter

danlu1:
Problem:

cfDNA samples are not released publicly.

Solution:
Make cfDNA samples public in main GENIE releases by removing sample_class_filter.

Testing:
Unit test has been added. And integration test has been done. Comparison report between `syn11611431.111` (with sample_class_filter) and `syn11611431.117` (without sample_class_filter) has been generated and got the expected result: two cfDNA sample rows are propagated.

Screenshot 2024-11-25 at 9 53 21 PM

### DIFF
--- a/genie/consortium_to_public.py
+++ b/genie/consortium_to_public.py
@@ -5,8 +5,13 @@ import os
 
 import pandas as pd
 import synapseutils
-from genie import (create_case_lists, database_to_staging, extract, load,
-                   process_functions)
+from genie import (
+    create_case_lists,
+    database_to_staging,
+    extract,
+    load,
+    process_functions,
+)
 
 logger = logging.getLogger(__name__)
 
@@ -148,7 +153,9 @@ def consortiumToPublic(
     allClin = clinicalDf[clinicalDf["SAMPLE_ID"].isin(publicReleaseSamples)]
     # check if cfDNA samples are present
     if not process_functions.check_values_in_column(allClin, "SAMPLE_CLASS", "cfDNA"):
-        logger.error("cfDNA samples should not be filtered out in the clinical dataframe.")
+        logger.error(
+            "cfDNA samples should not be filtered out in the clinical dataframe."
+        )
 
     allClin.to_csv(clinical_path, sep="\t", index=False)
 
--- a/genie/consortium_to_public.py
+++ b/genie/consortium_to_public.py
@@ -2,6 +2,7 @@
 
 import logging
 import os
+import sys
 
 import pandas as pd
 import synapseutils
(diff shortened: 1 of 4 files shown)


genie/consortium_to_public.py
```diff
@@ -147,7 +144,19 @@ def consortiumToPublic(
         query_string=f"SELECT * FROM {clinical_tier_release_scope_synid} where releaseScope = 'public'",
     )
 
+    # check if SAMPLE_CLASS is present
+    if not process_functions.check_values_in_column(
```
thomasyu888, line 152: These are great post processing checkers, eventually these will most likely be superceded with the new data quality framework rixing may want to put together.

thomasyu888 (commented):
🔥 LGTM! will leave for rixing for final review.

genie/consortium_to_public.py
```diff
@@ -147,7 +144,19 @@ def consortiumToPublic(
...
+        publicRelease, "fieldName", "SAMPLE_CLASS"
+    ):
+        logger.error("Must have SAMPLE_CLASS column in the public release scope.")
```
rxu17, line 155: Did this logging show up on the console log for a nextflow tower run for you? I think if not, we may have to do [something like this](https://github.com/Sage-Bionetworks/Genie/blob/develop/genie/extract.py#L16-L19) for everywhere we get `logging` to get it to show up in the console.

tests/test_process_functions.py
```diff
@@ -715,3 +714,74 @@ def test_that_create_missing_columns_returns_expected_output_with_multi_col_df()
...
+)
+def test_check_values_in_column_no_column(input_df, col, values):
+    with patch.object(process_functions, "logger") as mock_logger:
+        results = process_functions.check_values_in_column(input_df, col, values)
+    mock_logger.error.assert_called_once_with(
+        "Must have test_col column in the dataframe."
+    )
+
+
+@pytest.mark.parametrize(
```
rxu17, line 732: Could you add a test case for an empty dataframe? One that has the columns req in the check but no rows of data

rxu17 (approved):
LGTM! Just a couple of comments

genie/consortium_to_public.py
danlu1: The logging shows in the console when I run locally but I did't check it in tower. Let me check it.

danlu1: You are right. Thanks for catching this. The logging is not showing in the tower console or log file. I tweaked the code by adding stdout_handler.

tests/test_process_functions.py
danlu1: Done. It throws False as expected.

genie/consortium_to_public.py
danlu1: The logging info shows up in the Tower console now. 
Screenshot 2024-12-03 at 11 27 22 AM

PR #582, opened 2025-01-09
[GEN-1653] Add integration tests to GH actions workflow

rxu17:
**Purpose:** This adds automated integration tests for each step of the genie pipeline into our GH actions workflow. Still need to do the following:

- add conditions for when this should occur so it doesn't run every time there is a code change.
- `build-container` step should be pre-req to this step

**Testing:** 
Ran all testing pipeline steps here: https://github.com/Sage-Bionetworks/Genie/actions/runs/12662104568. It took close to an hour (mainly validation and consortium release steps taking the longest). Still need to look into why the case (separate ticket)

- [x] Verify pipeline steps don't have any underlying data changes with comparison script

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,9 @@ name: build
 on:
   push:
     branches: [main, develop, 'GEN*', 'gen*']
+    paths-ignore:
+      - '**.md'                # All Markdown files
+      - '**/docs/**'           # Documentation directory
 
   pull_request:
       types:
@@ -154,15 +157,6 @@ jobs:
         docker exec genie-container \
         python3 /root/Genie/bin/input_to_database.py main --project_id ${{ env.TEST_PROJECT_SYNID }} --deleteOld
 
-    - name: Run processing on mutation data in test pipeline (new MAF database creation on PR creation)
-      if: github.event_name == 'pull_request' && (github.event.action == 'opened' || github.event.action == 'reopened')
-      run: |
-        docker exec genie-container \
-        python3 /root/Genie/bin/input_to_database.py mutation --project_id ${{ env.TEST_PROJECT_SYNID }} \
-            --deleteOld \
-            --genie_annotation_pkg /root/annotation-tools \
-            --createNewMafDatabase
-
     - name: Run processing on mutation data in test pipeline
       if: github.event_name == 'push'
       run: |
@@ -170,6 +164,7 @@ jobs:
         python3 /root/Genie/bin/input_to_database.py mutation --project_id ${{ env.TEST_PROJECT_SYNID }} \
             --deleteOld \
             --genie_annotation_pkg /root/annotation-tools
+            --createNewMafDatabase
     
     - name: Run consortium release in test pipeline
       run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -144,6 +144,6 @@ jobs:
     
     - name: Run Tests in Docker Container
       run: |
-        docker run --rm ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ github.ref_name }}  \
+        docker run --rm ghcr.io/sage-bionetworks/${{ env.IMAGE_NAME }}:${{ github.ref_name }}  \
           python bin/input_to_database.py main --project_id syn7208886 --onlyValidate
   
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,8 @@ on:
 env:
   REGISTRY: ghcr.io
   IMAGE_NAME: ${{ github.repository }}
-  TEST_PROJECT_SYNID: syn7208886
+  TEST_PROJECT_SYNID: syn7208886        # synapse_id of the test synapse project
+  TEST_SEQ_DATE: Jul-2022               # SEQ_DATE to use for test pipeline. Should match the one used in nf-genie.
 
 jobs:
 
@@ -198,12 +199,12 @@ jobs:
     - name: Run consortium release in test pipeline
       run: |
         docker exec genie-container \
-        python3 /root/Genie/bin/database_to_staging.py Jan-2017 ../cbioportal TEST --test
+        python3 /root/Genie/bin/database_to_staging.py ${{ env.TEST_SEQ_DATE }} ../cbioportal TEST --test
 
     - name: Run public release in test pipeline
       run: |
         docker exec genie-container \
-        python3 /root/Genie/bin/consortium_to_public.py Jan-2017 ../cbioportal TEST --test
+        python3 /root/Genie/bin/consortium_to_public.py  ${{ env.TEST_SEQ_DATE }} ../cbioportal TEST --test
 
     - name: Stop and Remove Docker Container
       run: docker stop genie-container && docker rm genie-container
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,14 +37,23 @@ jobs:
       - name: Checkout Code
         uses: actions/checkout@v3
         with:
-          fetch-depth: 2
+          fetch-depth: 0
 
-      - name: Determine All Changes
-        id: all-changes
+      - name: Determine Changed Files
+        id: get-changes
         run: |
-          git fetch origin ${{ github.base_ref }}
-          ALL_CHANGED_FILES=$(git diff --name-only origin/${{ github.base_ref }} ${{ github.sha }})
-          echo "ALL_CHANGED_FILES=$ALL_CHANGED_FILES"
+          if [ "${{ github.event_name }}" == "push" ]; then
+            echo "Handling a push event..."
+            CURRENT_BRANCH=${{ github.ref_name }}
+            ALL_CHANGED_FILES=$(git diff --name-only HEAD^ HEAD)
+          elif [ "${{ github.event_name }}" == "pull_request" ]; then
+            echo "Handling a pull request event..."
+            BASE_BRANCH=${{ github.base_ref }}
+            HEAD_BRANCH=${{ github.head_ref }}
+            git fetch origin $BASE_BRANCH
+            ALL_CHANGED_FILES=$(git diff --name-only origin/$BASE_BRANCH HEAD)
+          fi
+          echo "All Changed Files: $ALL_CHANGED_FILES"
           echo "::set-output name=all-changes::$ALL_CHANGED_FILES"
 
       - name: Filter Non-Test Changes
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,9 @@ on:
     branches: [main, develop, 'GEN*', 'gen*']
 
   pull_request:
+      types:
+        - opened
+        - reopened
 
   release:
     types:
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -16,9 +24,36 @@ on:
...
+            ALL_CHANGED_FILES=$(git diff --name-only HEAD^ HEAD)
+          elif [ "${{ github.event_name }}" == "pull_request" ]; then
+            echo "Handling a pull request event..."
+            git fetch origin ${{ github.base_ref }}
+            ALL_CHANGED_FILES=$(git diff --name-only origin/${{ github.base_ref }} HEAD)
+          fi
+          echo "All Changed Files: $ALL_CHANGED_FILES"
+          NON_TEST_CHANGES=$(echo "$ALL_CHANGED_FILES" | grep -v '^tests/' || true)
+          echo "Non-Test Changes: $NON_TEST_CHANGES"
+          echo "::set-output name=non-test-changes::$NON_TEST_CHANGES"
```
thomasyu888, line 56: Just to get an understanding here. are you only running these tests when there are changes outside of the test folder?

.github/workflows/ci.yml
```diff
@@ -129,3 +144,87 @@ jobs:
...
+    runs-on: ubuntu-latest
+    if: ${{ needs.determine-changes.outputs.non-test-changes }}
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+
+    - name: Extract Branch Name
+      run: |
+        if [ "$GITHUB_HEAD_REF" != "" ]; then
+          echo "BRANCH_NAME=$GITHUB_HEAD_REF" >> $GITHUB_ENV
```
thomasyu888, line 160: Is there a set output command?

.github/workflows/ci.yml
```diff
@@ -129,3 +144,87 @@ jobs:
...
+    - name: Pull Public Docker Image from GHCR
+      run: |
+        docker pull ghcr.io/sage-bionetworks/genie:${{ env.BRANCH_NAME }}
+
+    - name: Start Docker Container
+      run: |
+        docker run -d --name genie-container \
+          -e SYNAPSE_AUTH_TOKEN="${{ secrets.SYNAPSE_AUTH_TOKEN }}" \
+          ghcr.io/sage-bionetworks/genie:${{ env.BRANCH_NAME }} \
+          sh -c "while true; do sleep 1; done"
```
thomasyu888, line 174: whats the motivation of this `while true` statement?

Whats your rationale for starting the container instead of just having 5 docker run commands?

thomasyu888 (approved):
🔥 Great work here, I'll let Dan be the final reviewer but minor comments

.github/workflows/ci.yml
rxu17: Yup, I didn't want the integration tests to run when there are only changes to the unit tests

thomasyu888: How would you differentiate between integration and unit tests being added?

.github/workflows/ci.yml
rxu17: Hmm from my understanding I thought that we had to keep the container "open" while running the code to keep it all in one container run/same state persistence. However thinking about it now, it should probably mirror nextflow where each step is probably separate docker run. 

I also wonder if this is the reason why I'm seeing differences in the outputs compared to outputs when I run the nextflow pipeline (something I am investigating ...)

.github/workflows/ci.yml
rxu17: > set output command
Oh like pull the docker image name instead from a previous job? I could do that but this seemed less complex

.github/workflows/ci.yml
rxu17: Hmmm our integration tests are going to be here as jobs in the github workflow while our unit tests are in the `tests` folder? That would be the differentiation

thomasyu888: Thanks for the reminder!  Does it make sense to include `tests` in the ignore paths?

.github/workflows/ci.yml
thomasyu888: Differences in outputs?!  What are the differences?

rxu17: All in the clinical consortium files. It's dropping more records for some reason (at least more compared to the nextflow pipeline run)

.github/workflows/ci.yml
rxu17: I don't think so because we want the `test` GH action job to run the unit tests in that folder

thomasyu888: You're right, I forgot the path exclude is for the entire workflow.

.github/workflows/ci.yml
```diff
@@ -6,8 +6,16 @@ name: build
 on:
   push:
     branches: [main, develop, 'GEN*', 'gen*']
+    paths-ignore:
+      - '**.md'                            # All Markdown files
+      - '**/docs/**'                       # Documentation directory
+      - '.github/workflows/codeql.yml'     # Code scanner
+      - '.github/workflows/build_docs.yml' # mkdocs GH workflow
```
danlu1, line 13: There are some paths I think we can ignore as well: [tests/testthat](https://github.com/Sage-Bionetworks/Genie/tree/develop/R/tests/testthat), and [.pre-commit-config.yaml](https://github.com/Sage-Bionetworks/Genie/blob/develop/.pre-commit-config.yaml). There maybe more. Since it's mainly about running through the pipeline, do you think we can use `path` instead to only point to related scripts folder?

.github/workflows/ci.yml
```diff
@@ -129,3 +144,87 @@ jobs:
...
+          echo "BRANCH_NAME=$GITHUB_HEAD_REF" >> $GITHUB_ENV
+        else
+          echo "BRANCH_NAME=${GITHUB_REF#refs/heads/}" >> $GITHUB_ENV
+        fi
+
+    - name: Pull Public Docker Image from GHCR
+      run: |
+        docker pull ghcr.io/sage-bionetworks/genie:${{ env.BRANCH_NAME }}
+
+    - name: Start Docker Container
```
danlu1, line 169: Just FYI, I saw https://docs.github.com/en/actions/writing-workflows/choosing-where-your-workflow-runs/running-jobs-in-a-container. It seems to be a way to spin up the container.

.github/workflows/ci.yml
rxu17: Yes so when I was looking at when integration tests should run - I found out it gets pretty extensive to use `paths` (because we also want to run integration tests when there are changes to non-script files like requirements.txt) so I chose to use `paths-ignore` and put paths that we typically run a lot and don't need integration tests to run. I can definitely put [.pre-commit-config.yaml](https://github.com/Sage-Bionetworks/Genie/blob/develop/.pre-commit-config.yaml) though but I don't think we really modify it.

So [this line](https://github.com/Sage-Bionetworks/Genie/pull/582/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR151) and the job it pulls from ignores any changes in the `tests` folder so that includes `tests/testthat`.

.github/workflows/ci.yml
rxu17: Thanks I think I saw that. I'll take a look once I resolve the data discrepancies.

.github/workflows/ci.yml
rxu17: It was due to the different seq_dates used :')

.github/workflows/ci.yml
rxu17: Thanks for linking this! 

It looks like the github integrated container is much faster/just not using `docker exec` is the key. After some research, looks like `docker exec` causes extra latency and overhead and repeated calls builds up and just increase the slowness. Known issue: https://github.com/docker/for-linux/issues/466

rxu17: Been trying out the github integrated container that you linked and it gives [similar run times](https://github.com/Sage-Bionetworks/Genie/actions/runs/13154694462) to the current layout here after a couple of runs (ignore the error message)

After some research, I think the bottleneck is just that Github actions only has 2 cores of CPU to work with and our pipeline runs even with the test pipeline seem heavy duty. Our local macs + ec2 containers just have much higher processing speeds because of the CPU available. It seems overkill right now to have a self-hosted runner and spin up an ec2 instance for it to run the integration tests on (especially cost, and also we want to be careful about exposing access). I'll keep looking into it. However, asking us to wait a varying 30 min to 1hr+ for integration tests to run can be pretty painful.

An alternative we can consider is just it executing a nextflow run (and hoping it captures the error so the GH run is labeled as FAILED). Downside is that nextflow runs don't offer the full comprehensive logging (due to visibility issue with our logging in nextflow) that running each step here provides especially if each step succeeds. However we currently use the nextflow runs' logging to troubleshoot our consortium releases so I don't see this being a huge downside. I do see this: [seqeralabs/action-tower-launch](https://github.com/marketplace/actions/action-tower-launch) which looks very promising/okay difficulty to integrate. Will prototype with it and see how well it works.

rxu17: After attempting to setup up the above action, I think there's a lot more to be done to auto-integrate in the nextflow run (like how to pull in the nextflow compute env programmatically since it changes via github actions) or debugging it (the error messages for the action before you even have a run going are not helpful because there are none). Will add my findings to the relevant JIRA ticket that has us integrate in a nextflow run.

danlu1: > Been trying out the github integrated container that you linked and it gives [similar run times](https://github.com/Sage-Bionetworks/Genie/actions/runs/13154694462) to the current layout here after a couple of runs (ignore the error message)

I'm not sure how to check the run time since the attached one is failed. I saw one [succeed job](https://github.com/Sage-Bionetworks/Genie/actions/runs/13151347465) takes 1.5hr which is a bit of concern.

danlu1: Yes, using Tower can at least avoid the CPU limitation in GH action. We might be able to use compute_env to specify desired env and [actions/upload-artifact](https://github.com/actions/upload-artifact) to save verbose log somewhere.

rxu17: Yeah so here are the previous integration test run times: https://github.com/Sage-Bionetworks/Genie/actions?query=workflow%3Abuild+branch%3Agen-1653-add-integration-tests+is%3Asuccess

If you click on the jobs, you'll see that for each step run, it is random which step takes a lot of time. In the failed run, you can see that just the processing of non-mutation data takes 30+ min. I ran it previously, and it was also taking a while as well. I can definitely try for more runs but I think after about 3 runs, it seems to be a CPU overhead issue.

rxu17: Yeah so since our nextflow compute_env always gets updated, we have to find a way to pull it programmatically  in GH actions (it's not very self-explanatory). Yes [actions/upload-artifact](https://github.com/actions/upload-artifact) starts working once the nextflow run launch is successful and kicks in. There's a lot of logging we can do with [seqeralabs/action-tower-launch](https://github.com/marketplace/actions/action-tower-launch) once we are able to get it setup.

rxu17: I think given the long runtimes, if it's a big concern. I can just close this PR and mark as `Won't Do`, and we can escalate this one: https://sagebionetworks.jira.com/browse/GEN-1658. But I'm worried about the limitation of GH action's ability to pull dynamic values from our tower env like the compute_env that would end up being a blocker there.

It's also odd because the initial test runs of the integration testing were only 30 min for the whole pipeline. I'll monitor another run: https://github.com/Sage-Bionetworks/Genie/actions/runs/13170477849 using the github's container workflow.

thomasyu888: Since you've done quite a bit of work here already, I'm happy to merge this first and we can iterate on the solution. 

Sure these are long test times, but it does give us a starting point. 

GitHub runners are inconsistent in general. I've seen this with all of our jobs that take longer than 10 minutes to run.

PR #587, opened 2025-03-10
[GEN-1332] add sv to gene panel matrix

danlu1:
Problem:

There is a request to add SV column in the current gene panel matrix file

Solution:
1. Output sv_sample list in store_sv_files function
2. Add the sv column in the store_data_gene_matrix function

Testing:
Unit tests have been added and integration test has been done on my local.

### DIFF
--- a/genie/database_to_staging.py
+++ b/genie/database_to_staging.py
@@ -877,6 +877,9 @@ def store_sv_files(
         keep_for_merged_consortium_samples: Samples to keep for merged file
         current_release_staging: Staging flag
         center_mappingdf: Center mapping dataframe
+
+    Returns:
+        List of SV Samples
     """
     logger.info("MERING, FILTERING, STORING SV FILES")
     sv_df = extract.get_syntabledf(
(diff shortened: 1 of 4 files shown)


genie/database_to_staging.py
```diff
@@ -1613,6 +1619,15 @@ def store_data_gene_matrix(
     wes_panel_cna = data_gene_matrix["cna"].isin(wes_seqassayids)
     data_gene_matrix = data_gene_matrix[~wes_panel_cna]
 
+    # Add SV column into gene panel file
```
rxu17, line 1620: Nit: I think this could be a reusable function that you can apply to both cna and sv columns? It can also reduce the complexity of the testing for this function including the amt of mocking on top of reducing code redundancy

tests/test_database_to_staging.py
```diff
@@ -886,3 +885,143 @@ def test_store_mutation_in_cis_files_to_staging_no_centers(
...
+            filepath=os.path.join(database_to_staging.GENIE_RELEASE_DIR, "data_sv.txt"),
+            parentid="syn123",
+            version_comment="TESTING",
+            name="data_sv.txt",
+            used="syn5678.1",
+        )
+        assert sv_sample == ["GENIE-1", "GENIE-3"]
+
+
+def test_store_data_gene_matrix(syn):
```
rxu17: Maybe other test case(s) where nothing should get filtered out in the data gene matrix in the `sv` columns / no wes panels ids are present in current data for example

rxu17 (approved):
LGTM! Just a couple of comments

genie/database_to_staging.py
danlu1: Sure. I will tweak this.

PR #588, opened 2025-03-11
[GEN-1616] Automate update of Genome Nexus truststore

rxu17:
### **Problem** 
The GENIE pipeline depends on a truststore file that was provided by Genome Nexus team and this file would go out of date once every few months. Currently this update is done [manually via this procedure](https://sagebionetworks.jira.com/wiki/spaces/APGD/pages/3016687662/Updating+Genome+Nexus+Annotator+and+Dependencies#Updating-the-trust-ssl-file). 

The issue with manual is that we would get an [error that indicates that all mutation variants got annotated to be `FAILED`](https://sagebionetworks.jira.com/wiki/spaces/APGD/pages/2856943651/GENIE+Pipeline+Troubleshooting#Issue:-Genome-Nexus-annotator-produces-all-failed-annotations), and that was an indicator that the trust file needed updating but often that means we have to go through the entirety of mutation processing (~2 days) before this error gets triggered.

### **Solution** 
This PR will do the following:
1. Update the latest truststore file present [here](https://www.synapse.org/Synapse:syn22053204) - this is scheduled to run every day
2. Run the mutation processing and consortium release steps (should pass both steps without error) on the test pipeline to make sure the update was successful. 

This way we don't have to worry about running into an outdated trustfile while in the middle of processing given the frequency of updates.

### **Testing**
1. Successful test build (update and checking that the update succeeded) is here: https://github.com/Sage-Bionetworks/Genie/actions/runs/13798855778
2. See file update success [here](https://www.synapse.org/Synapse:syn22053204)

### DIFF
--- a/.github/workflows/automate_truststore.yml
+++ b/.github/workflows/automate_truststore.yml
@@ -1,6 +1,7 @@
 name: Auto-update Genome Nexus Truststore file
 
 on:
+  push:
   schedule:
     - cron: "0 3 * * 1-5" # Runs at 3 AM UTC Monday to Friday
   workflow_dispatch:
--- a/.github/workflows/automate_truststore.yml
+++ b/.github/workflows/automate_truststore.yml
@@ -1,7 +1,6 @@
 name: Auto-update Genome Nexus Truststore file
 
 on:
-  push:
   schedule:
     - cron: "0 3 * * 1-5" # Runs at 3 AM UTC Monday to Friday
   workflow_dispatch:
--- a/.github/workflows/automate_truststore.yml
+++ b/.github/workflows/automate_truststore.yml
@@ -2,7 +2,7 @@ name: Auto-update Genome Nexus Truststore file
 
 on:
   schedule:
-    - cron: "0 3 * * 1-5" # Runs at 3 AM UTC Monday to Friday
+    - cron: "0 3 * * 2,5" # Runs at 3 AM UTC on Tuesday (2) and Friday (5)
   workflow_dispatch:
 
 env:
--- a/.github/workflows/automate_truststore.yml
+++ b/.github/workflows/automate_truststore.yml
@@ -9,7 +9,7 @@ on:
 env:
     PROD_DOCKER: ghcr.io/sage-bionetworks/genie:main
     SYNAPSE_AUTH_TOKEN: ${{ secrets.SYNAPSE_AUTH_TOKEN }}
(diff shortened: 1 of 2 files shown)


thomasyu888 (commented):
🔥 awesome work, just some questions: Does the truststore need to be updated daily?  What happens when the truststore is updated mid maf processing?

rxu17:
> Does the truststore need to be updated daily?

I guess I put daily so it doesn't miss a day/less risk of having outdated file because I honestly don't know when they update the truststore (very random) but it could be less infrequent. Maybe I could just have it run right before every consortium release (that's a thought) every month but since we also have main genie development that uses the same truststore, it may be better to do it more frequently.

Any thoughts on frequency there?

> What happens when the truststore is updated mid maf processing?

So since we pull the truststore at the beginning of maf processing, it would just use the truststore version that was at the beginning of processing. Since the actual update is quick, I don't see this being an issue? It's interesting because even if there isn't an actual "update", the command will still pass and say updated

danlu1 (approved):
LGTM. Just a qq. Is the truststore only saved in TESTING project even for prod processing? I understand that the SynID of TESTING project is stated in the SOP.

thomasyu888:
> So since we pull the truststore at the beginning of maf processing, it would just use the truststore version that was at the beginning of processing. Since the actual update is quick, I don't see this being an issue? It's interesting because even if there isn't an actual "update", the command will still pass and say updated

My concern is that maf processing can be 2 days long for some sites, so let's take example

1. processing of vcfs starts for a site
2. Truststore is updated
3. Maf processing

I forget when the trust store file is pulled down in the maf processing.

rxu17:
> LGTM. Just a qq. Is the truststore only saved in TESTING project even for prod processing? I understand that the SynID of TESTING project is stated in the SOP.

Yes, we could just migrate it to production project right now similar to how the oncotree code url is in production folder but gets used by all projects. Since there's technically no `dev` version of the annotator and the trustfile/dep, it's like a variable we pull from.

rxu17:
> My concern is that maf processing can be 2 days long for some sites, so let's take example

    processing of vcfs starts for a site
    Truststore is updated
    Maf processing

> I forget when the trust store file is pulled down in the maf processing.

So let's say we update it every day. So the processing of mutation data would use the truststore version from the previous day (that should be fine, highly unlikely genome nexus team updates the server after that but before we start maf processing). The trust store file gets pulled at the beginning of maf processing. Does that help? Not sure the concern. I do think maybe the current cadence is too high and we could reduce it to something like twice a week (Tues, Fri)

PR #594, opened 2025-05-01
[GEN-1890,GEN-869] Redact all age columns

danlu1:
# **Problem:**

Currently, 

1. BIRTH_YEAR is always redacted when the difference between BIRTH_YEAR and YEAR_CONTACT or YEAR_DEATH exceeds 89 years (or 32,485 days) or less than 18 years AND when AGE_AT_SEQ_REPORT, INT_CONTACT, INT_DOD greater than 32485 or less than 6570 or have < or > in the value.
2. YEAR_CONTACT and YEAR_DEATH are never redacted. 
3. AGE_AT_SEQ_REPORT, INT_CONTACT, INT_DOD are redacted when they are greater than 32485 or less than 6570 or have < or > in the value.

However, we want to all age variables are redacted if ONE variable in a sample/patient is flagged to be over 89.

# **Solution:**

Bind the boolean vector indicating which field to be redacted for all age column checks and replace actual values with redacted label: `cannotReleaseHIPAA` and `withheld`.

# **Testing:**

Unit tests have been added.

### DIFF
--- a/genie/database_to_staging.py
+++ b/genie/database_to_staging.py
@@ -208,6 +208,8 @@ def redact_phi(
     to_redact_year_death = _to_redact_difference(
         clinicaldf["BIRTH_YEAR"], clinicaldf["YEAR_DEATH"]
     )
+
+    # get complete index list to redact
     complete_redact = (
         complete_redact
         | to_redact_birth_year
@@ -217,6 +219,16 @@ def redact_phi(
     # redact all age columns for >89
     clinicaldf.loc[
         complete_redact,
+        ["BIRTH_YEAR", "YEAR_CONTACT", "YEAR_DEATH"] + interval_cols_to_redact,
+    ] = "cannotReleaseHIPAA"
+
+    # get all impacted patient rows
+    patient_ids = clinicaldf.loc[
+        clinicaldf["BIRTH_YEAR"] == "cannotReleaseHIPAA", "PATIENT_ID"
+    ]
+
+    clinicaldf.loc[
+        clinicaldf["PATIENT_ID"].isin(patient_ids),
         ["BIRTH_YEAR", "YEAR_CONTACT", "YEAR_DEATH"] + interval_cols_to_redact,
     ] = "cannotReleaseHIPAA"
 
--- a/genie/database_to_staging.py
+++ b/genie/database_to_staging.py
@@ -159,14 +159,16 @@ def _to_redact_difference(df_col_year1, df_col_year2):
         df_col_year2: Dataframe column/pandas.Series of a year column
 
     Returns:
-        pandas.Series: to redact boolean vector
+        tuple: pandas.Series: to redact boolean vector
+               pandas.Series: to redact pediatric boolean vector
 
     """
     # Add in errors='coerce' to turn strings into NaN
     year1 = pd.to_numeric(df_col_year1, errors="coerce")
     year2 = pd.to_numeric(df_col_year2, errors="coerce")
     to_redact = year2 - year1 > 89
-    return to_redact
+    to_redact_ped = year2 - year1 < 18
+    return to_redact, to_redact_ped
 
 
 # TODO: Add to transform.py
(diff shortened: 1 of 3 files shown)


danlu1:
The failed unit tests are the old test in test_filters. I will update them. @thomasyu888 question. Do you know why you put the unit test in the [test_filters](https://github.com/Sage-Bionetworks/Genie/blob/d70ed8729b79eb506dad6e797d09a05128bed864/tests/test_filters.py#L129C1-L181C1) file instead of test_database_to_staging where the code blocks belong to?

genie/database_to_staging.py
```diff
@@ -152,21 +152,23 @@ def _redact_year(df_col):
...
+        tuple: pandas.Series: to redact boolean vector
+               pandas.Series: to redact pediatric boolean vector
 
     """
     # Add in errors='coerce' to turn strings into NaN
     year1 = pd.to_numeric(df_col_year1, errors="coerce")
     year2 = pd.to_numeric(df_col_year2, errors="coerce")
     to_redact = year2 - year1 > 89
-    return to_redact
+    to_redact_ped = year2 - year1 < 18
```
thomasyu888: Ah! Don't redact <18. This ticket was only for >89 because we have plans on removing the redaction plan for <18.

I don't know if that complicates things if we keep current behavior for <18

thomasyu888 (changes requested):

genie/database_to_staging.py
danlu1: ahh I thought we want to redact ped anyway but just don't know how to label it. Got it. I will remove it

genie/database_to_staging.py
```diff
@@ -189,23 +191,47 @@ def redact_phi(
...
+    to_redact_year_contact, to_redact_year_contact_ped = _to_redact_difference(
         clinicaldf["BIRTH_YEAR"], clinicaldf["YEAR_CONTACT"]
     )
-    clinicaldf.loc[to_redact, "BIRTH_YEAR"] = "cannotReleaseHIPAA"
-    to_redact = _to_redact_difference(
+    to_redact_year_death, to_redact_year_death_ped = _to_redact_difference(
         clinicaldf["BIRTH_YEAR"], clinicaldf["YEAR_DEATH"]
     )
-    clinicaldf.loc[to_redact, "BIRTH_YEAR"] = "cannotReleaseHIPAA"
+    complete_redact = (
```
thomasyu888, line 213: Does this create a tuple of True/False?

danlu1: This creates a pd.Series of True/False.

thomasyu888:
> The failed unit tests are the old test in test_filters. I will update them. @thomasyu888 question. Do you know why you put the unit test in the [test_filters](https://github.com/Sage-Bionetworks/Genie/blob/d70ed8729b79eb506dad6e797d09a05128bed864/tests/test_filters.py#L129C1-L181C1) file instead of test_database_to_staging where the code blocks belong to?

I'm not entirely sure - i probably wanted to separate out the filters into a filters module and never got around to it. Feel free to merge them.

genie/database_to_staging.py
```diff
@@ -159,13 +159,14 @@ def _to_redact_difference(df_col_year1, df_col_year2):
         df_col_year2: Dataframe column/pandas.Series of a year column
 
     Returns:
-        pandas.Series: to redact boolean vector
+        tuple: pandas.Series: to redact boolean vector
```
thomasyu888: nit: I think this is redundant?

thomasyu888 (approved):
🔥 LGTM! This is fantastic work - just one nit

genie/database_to_staging.py
```diff
@@ -189,23 +190,35 @@ def redact_phi(
...
     # years would change every single year. (e.g. <1926, >1998 would be
     # inaccurate every year)
+    complete_redact = pd.Series([False] * len(clinicaldf))
     for col in interval_cols_to_redact:
-        to_redact, to_redactpeds = _to_redact_interval(clinicaldf[col])
-        clinicaldf.loc[to_redact, "BIRTH_YEAR"] = "cannotReleaseHIPAA"
-        clinicaldf.loc[to_redact, col] = ">32485"
-        clinicaldf.loc[to_redactpeds, "BIRTH_YEAR"] = "withheld"
-        clinicaldf.loc[to_redactpeds, col] = "<6570"
+        to_redact, to_redact_peds = _to_redact_interval(clinicaldf[col])
```
thomasyu888, line 195: This complicates things a little bit: https://sagebionetworks.slack.com/archives/C0350GLJ7M4/p1746489087447869?thread_ts=1746483192.246719&cid=C0350GLJ7M4. 

Basically if a patient contains fields that are to be redacted, every age variable that that patient across all samples need to be redacted

thomasyu888 (changes requested):

genie/database_to_staging.py
danlu1: Good catch. This should be removed since we don't generate vector for ped.

tests/test_database_to_staging.py
```diff
@@ -1045,3 +1046,608 @@ def test_store_data_gene_matrix(syn, wes_seqassayids, expected_output):
...
+                    ],  # in years: 52*365+1, 54*365+2, 20*365, 89*365
+                    "Other_col": ["a", "b", "c", "d"],
+                }
+            ),
+        },
+        {
+            "name": "has_column_to_be_redacted",
+            "input_df": pd.DataFrame(
+                {
+                    "PATIENT_ID": ["a", "b", "c", "d", "e", "e"],  # redact b, d, e
```
danlu1, line 1288: @thomasyu888 this is a test case for a patient that to be redacted has multiple samples.

thomasyu888 (approved):
🔥 LGTM!  Thanks for adding this.  Lets wait for the PMO meeting tomorrow to determine whether we should cherry pick this into the `main` branch.

genie/database_to_staging.py
```diff
@@ -189,23 +190,47 @@ def redact_phi(
     # Moved to cannotReleaseHIPAA and withheld because the HIPAA
     # years would change every single year. (e.g. <1926, >1998 would be
     # inaccurate every year)
+    complete_redact = pd.Series([False] * len(clinicaldf))
     for col in interval_cols_to_redact:
-        to_redact, to_redactpeds = _to_redact_interval(clinicaldf[col])
```
thomasyu888, line 193: To move forward

```
    to_redact = (col_int > phi_cutoff) | contain_greaterthan
```

This is of concern... so lets talk through this in a meeting, BUT...

89*365 = 32485

AND when you are 32485 days old, you are 89.  But if we are looking for > 32485, we are actually including the first day of 89.  So... should we be filtering out anything >=32485


@Chelsea-Na / @danlu1

danlu1: Confirmed with xindi that we don't use `equal to` check: https://sagebionetworks.jira.com/browse/GEN-1890?focusedCommentId=248144

PR #595, opened 2025-05-14
[GEN-1953] save maf table to prod

danlu1:
# **Problem:**

During MAF processing, we generate a new MAF table and attempt to move it to the GENIE Testing project, which has more limited permissions. However, due to the new permission restrictions, this operation fails when transferring the entity from a project with more restrictive access to one with less restrictive access.

# **Solution:**

Update parent ID for table saving section.

### DIFF
--- a/bin/input_to_database.py
+++ b/bin/input_to_database.py
@@ -6,13 +6,8 @@ import argparse
 import logging
 from datetime import date
 
-from genie import (
-    config,
-    extract,
-    input_to_database,
-    process_functions,
-    write_invalid_reasons,
-)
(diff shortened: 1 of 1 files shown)


thomasyu888 (approved):
🔥 LGTM!

bin/input_to_database.py
```diff
@@ -3,16 +3,11 @@
 
 """
 import argparse
-from datetime import date
 import logging
+from datetime import date
 
-from genie import (
```
thomasyu888, line 9: This is auto formatted by black I think, so I would keep it as is.

PR #597, opened 2025-06-10
[GEN-2027] Add staging for mutation in cis

rxu17:
# **Problem:**
Staging is currently run with the [mutation in cis process](https://sagebionetworks.jira.com/wiki/spaces/APGD/pages/3976691723/Mutation+in+CIS+Process+WIP) disabled. This creates a discrepancy when comparing staging and production runs on the same data as production runs will often have a lot more variants filtered out compared to staging. 

See [background process](https://sagebionetworks.jira.com/wiki/spaces/APGD/pages/3976691723/Mutation+in+CIS+Process+WIP)

# **Solution:**
Staging is enabled when running mutation in cis process. 

# **Testing:**
- Unit tests are written to test that staging is called as expected
- Ran on testing pipeline to ensure no changes are present
- Real test is after merging, running staging on `develop` branch

### DIFF
--- a/R/mergeCheck.R
+++ b/R/mergeCheck.R
@@ -9,8 +9,8 @@ parser$add_argument("--staging",
                     action = "store_true",
                     help = "Use staging files")
 args <- parser$parse_args()
-testing <- args$testing
-staging <- args$staging
+testing <- as.logical(args$testing)
+staging <- as.logical(args$staging)
 
 library(synapser)
 library(VariantAnnotation)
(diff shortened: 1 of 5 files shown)


R/mergeCheck.R
```diff
@@ -39,13 +39,10 @@ variant_limit = 100000
...
 
 #testing = as.logical(args[1])
-if (testing) {
-  databaseSynIdMappingId = 'syn11600968'
-} else if (staging) {
-  databaseSynIdMappingId = 'syn12094210'
-} else {
-  databaseSynIdMappingId = 'syn10967259'
-}
+databaseSynIdMappingId <- get_database_to_synapse_mapping_synid(
```
thomasyu888, line 42: Nit: Thanks for adding this into a function, I would say that this is actually a hack and should probably use project_id to obtain the dataBaseSynIdMapping.  However, that probably requires a whole configuration change across the board so I'd leave things as is.

thomasyu888 (approved):
🔥 LGTM!  I'll defer to @danlu1 for final review

danlu1 (approved):
LGTM

PR #599, opened 2025-07-22
[GEN-1789, GEN-1775] Update new ubuntu version to support python 3.10, python 3.11

thomasyu888:
# **Problem:**

We are using the an older version of ubuntu (focal 20.04 which is no [longer actively supported](https://ubuntu.com/20-04)) and we need to upgrade the version.

# **Solution:**
Switch to using ubuntu 22.04 (jammy), which has support for our current R version 4.3.3 and for the `python3.10-venv`
- We also had to upgrade the virtualenv used for our R scripts to be python 3.8 -> python 3.10 (as jammy doesn't have python 3.8 anymore). I believe we can't use noble as noble doesn't have support for our R version. 
- Another caveat of updating is we have to lock in the reticulate virtualenv's pandas and numpy versions to be the stable versions `pandas <=2.03 and numpy<=1.24.4` as currently there isn't support for working with `pandas > 2.0.3 and numpy > 1.24.4` for reticulate ([see investigation into this issue that when converting to pandas df in reticulate, the columns change to list types](https://sagebionetworks.jira.com/browse/GEN-1775?focusedCommentId=261298)) even in [`synapser`](https://github.com/Sage-Bionetworks/synapser/blob/v2.1.1-rc/tools/installPythonClient.R).
- As part of ensuring the build successfully completes, we needed to pin down the specific `opentelemetry-semantic-conventions` version

# **Testing:**
- Unit tests run
- Integration tests ran all the way through
- Comparison tests on testing pipeline's results between this branch's docker image and the `develop` docker image pull up no differences


rxu17:
@thomasyu888 requesting review

thomasyu888 (commented):
🔥 LGTM!

Dockerfile
```diff
@@ -24,8 +24,8 @@ RUN apt-get update && apt-get install -y --allow-unauthenticated --no-install-re
 		python3-pip \
 		python3-dev \
 		git \
-		r-base-core=4.3.3-1.2004.0 \
-		r-base-dev=4.3.3-1.2004.0 \
+		r-base-core=4.3.3-1.2204.0 \
+		r-base-dev=4.3.3-1.2204.0 \
```
danlu1, line 28: Just out of curiosity, how did you decide the version here?

danlu1 (approved):
LGTM

Dockerfile
rxu17: It's by the name of the ubuntu version we use for the dockerfile. Previously it was ubuntu focal which was 20.04, but now it's ubuntu jammy so it's 22.04.

PR #601, opened 2025-07-24
[GEN-998] Resolve fill na issue

rxu17:
# **Problem:**
There was a failing test `test_append__append_rows` in `tests/test_process_functions.py` due to fillna("", inplace=True) due to [inconsistent behavior coercing into numeric dtype vs not numeri dtype](https://sagebionetworks.jira.com/browse/GEN-978). However this appears to not be an issue anymore (no more failing test)
JIRA Ticket: https://sagebionetworks.jira.com/browse/GEN-998

Depends on #599 

# **Solution:**
However after running this on the recent build of the pandas branch `gen-178-update-pandas-2`, the tests **pass** when running in python 3.10 and python 3.11. This doesn't appear to be an issue anymore (previous it was failing with the tests when running in 3.8/3.9). 

# **Testing:**
- Unit tests all pass
- Tests pass when running in docker image


thomasyu888 (approved):
🔥 LGTM!

PR #602, opened 2025-08-01
[GEN 1262] Enable dashboard generation for testing mode

danlu1:
# **Problem:**

Currently, there’s no way to test updated dashboard code against the test version of the pipeline. The only time we can verify that our dashboard code works is when it runs on a staging or production consortium release.

To improve this, we should update the dashboard generator code to support running on the test data and pipeline as fully as possible. This would allow us to catch issues earlier and reduce dependency on production releases for validation.

# **Solution:**
1. Enable dashboard generator to pull test data
2. Add the ability to turn on dashboard generator for test pipeline
# **Testing:**
The script was run through by calling 
```
Rscript R/dashboard_markdown_generator.R TESTING --template_path templates/dashboardTemplate.Rmd --testing
```
and the output html is saved to the [expected folder](https://www.synapse.org/Synapse:syn12299959).

### DIFF
--- a/bin/consortium_to_public.py
+++ b/bin/consortium_to_public.py
@@ -20,7 +20,9 @@ PWD = os.path.dirname(os.path.abspath(__file__))
 
 
 # TODO: Move to genie.database_to_staging.py
-def generate_dashboard_html(genie_version: str, staging=False, testing=False):
+def generate_dashboard_html(
+    genie_version: str, staging: bool = False, testing: bool = False
+):
     """Generates dashboard html writeout that gets uploaded to the
     release folder
 
--- a/bin/consortium_to_public.py
+++ b/bin/consortium_to_public.py
@@ -20,7 +20,7 @@ PWD = os.path.dirname(os.path.abspath(__file__))
 
 
 # TODO: Move to genie.database_to_staging.py
-def generate_dashboard_html(genie_version, staging=False, testing=False):
+def generate_dashboard_html(genie_version: str, staging=False, testing=False):
     """Generates dashboard html writeout that gets uploaded to the
     release folder
 
--- a/bin/consortium_to_public.py
+++ b/bin/consortium_to_public.py
@@ -18,7 +18,7 @@ PWD = os.path.dirname(os.path.abspath(__file__))
 
 
 # TODO: Move to genie.database_to_staging.py
-def generate_dashboard_html(genie_version, staging=False):
+def generate_dashboard_html(genie_version, staging=False, testing=False):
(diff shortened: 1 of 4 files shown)


bin/consortium_to_public.py
```diff
@@ -254,7 +258,7 @@ def main(args):
         "--staging", action="store_true", help="Store into staging folder"
     )
 
-    parser.add_argument("--test", action="store_true", help="Store into staging folder")
+    parser.add_argument("--test", action="store_true", help="Store into testing folder")
```
danlu1, line 268: I think this is a typo from the original code

rxu17 (approved):
LGTM! Just a couple of minor comments

bin/database_to_staging.py
```diff
@@ -86,7 +86,7 @@
 PWD = os.path.dirname(os.path.abspath(__file__))
 
 
-def generate_dashboard_html(genie_version, staging=False):
+def generate_dashboard_html(genie_version, staging=False, testing=False):
```
rxu17: Update docstring with new param

bin/consortium_to_public.py
```diff
@@ -18,7 +18,7 @@
 
 
 # TODO: Move to genie.database_to_staging.py
-def generate_dashboard_html(genie_version, staging=False):
+def generate_dashboard_html(genie_version, staging=False, testing=False):
```
rxu17: Update docstring with new param

bin/consortium_to_public.py
```diff
@@ -2,29 +2,32 @@
...
+)
 
 logger = logging.getLogger(__name__)
 
 PWD = os.path.dirname(os.path.abspath(__file__))
 
 
 # TODO: Move to genie.database_to_staging.py
-def generate_dashboard_html(genie_version, staging=False):
+def generate_dashboard_html(genie_version, staging=False, testing=False):
```
rxu17: nit: If you want to get past the failing sonarcloud rating thing, would add type hints to this

bin/database_to_staging.py
```diff
@@ -81,18 +79,22 @@
...
+# import time
+
+
 logger = logging.getLogger(__name__)
 
 PWD = os.path.dirname(os.path.abspath(__file__))
 
 
-def generate_dashboard_html(genie_version, staging=False):
+def generate_dashboard_html(genie_version, staging=False, testing=False):
```
rxu17: nit: If you want to get past the failing sonarcloud rating thing, would add type hints to this

rxu17 (approved):
LGTM! Would just make sure the integration tests and then should be good

PR #604, opened 2025-08-13
[GEN-1892] unmask pediatric data

danlu1:
# **Problem:**

We want to unmask pediatric patient data in main GENIE and have them be available for case selection in BPC.

# **Solution:**

Remove the filter to allow actual age values/granular year to be released. 

# **Testing:**
Unit test cases have been added.


tests/test_database_to_staging.py
```diff
@@ -1063,3 +1064,608 @@ def test_store_data_gene_matrix(syn, wes_seqassayids, expected_output):
...
+                    "Unknown",
+                    "Not Applicable",
+                    np.nan,
+                ]  # in years: ">89", "<18"
+            ),
+            pd.Series([True, False, False, False, False, False]),
+            pd.Series([False, True, False, False, False, False]),
+        ),
+    ],
+    ids=["no_cutoff_value", "has_cutoff_value", "non_numeric_values", "has_NAs"],
```
rxu17: nit: I would add your cutoff value 33215 from your first test case to your 2nd test case since it fits under the 2nd case's id `has_cutoff_value`

rxu17 (approved):
LGTM, just a small comment!

tests/test_database_to_staging.py
danlu1: The cutoff values here mean: 18 and 89 which are 6570 and 32485.

rxu17: Gotcha, my head is thinking that cutoff values are anything that hits the threshold/cutoff point and beyond

rxu17: nit: maybe for clarity, it might make sense to also have tests that are clear in what kind of redaction is there (since these functions are testing redaction) ? For example have a test with all values redacted and you have a mix of strings - actual non-numeric and numeric. Then have a test with no values redacted.

danlu1: Updated the test case names.

PR #606, opened 2025-08-15
[GEN-2222] Fix failing timezone tests by using utc timezone

rxu17:
# **Problem:**
There are now failing tests that involve the `to_unix_epoch_time` function from synapseclient (used when we are trying to save the [modifiedOn timestamp for files like validation status synapse table](https://www.synapse.org/Synapse:syn8523326). 

The converted UNIX time doesn't match when running locally on mac vs running on docker image/github actions. This is because we upgraded the `synapseclient` from `2.7.2` to `4.9.0` where the client code changed for that function from [2.7.2](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/v2.7.2/synapseclient/core/utils.py#L380-L388) to [now](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/v4.9.0/synapseclient/core/utils.py#L679-L697) and it’s actually doing what it is supposed to do now - use local system time when doing the conversion when running on mac (PDT) vs docker time (UTC).

JIRA Ticket: https://sagebionetworks.jira.com/browse/GEN-2222

# **Solution:**
Wanted to have a discussion about this solution and if we care about making sure if tracking if we are running in PDT timezone or UTC timezone. For consistency's sake, I feel like we want to use one timezone (even though we use nextflow platform for all runs, just in case we run things locally) so I opted to convert from local to UTC for all scenarios before using the actual conversion to UNIX time. 

Other option is to just force the timezone to be UTC just for running the tests but we could have different time zones when uploading to synapse.

UPDATE: Make decision to convert to UTC timezone because there will be an obvious discrepancy in the modified on dates, see testing report for details

# **Testing:**
- [X] Unit Tests pass
- [X] Integration tests pass
- [X] Check that modifiedOn in validation status table match (after validation) when running locally on mac (local time) vs the integration tests (which is run in the docker using UTC time) and compare that to a run locally on mac (using the updated synapseclient version of 4.9 but **without** the fix.  Since the files haven't changed themselves, it should hopefully produce the same result.

Report is available here: https://sagebionetworks.jira.com/browse/GEN-2222?focusedCommentId=263423

### DIFF
--- a/genie/dashboard_table_updater.py
+++ b/genie/dashboard_table_updater.py
@@ -7,7 +7,6 @@ import os
 
 import pandas as pd
 import synapseclient
-from synapseclient.core.utils import to_unix_epoch_time
 
 from genie import extract, load, process_functions
 
@@ -570,7 +569,7 @@ def string_to_unix_epoch_time_milliseconds(string_time):
     datetime_obj = datetime.datetime.strptime(
         string_time.split(".")[0], "%Y-%m-%dT%H:%M:%S"
     )
-    return to_unix_epoch_time(datetime_obj)
+    return process_functions.to_unix_epoch_time_utc(datetime_obj)
 
 
 def update_data_release_file_table(syn, database_mappingdf):
--- a/genie/dashboard_table_updater.py
+++ b/genie/dashboard_table_updater.py
@@ -555,7 +555,7 @@ def update_wiki(syn, database_mappingdf):
     syn.store(wikipage)
 
 
-def string_to_unix_epoch_time_milliseconds(string_time):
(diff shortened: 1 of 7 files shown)


danlu1:
I would vote for converting to UTC before using the actual conversion to UNIX time to have a consistency timezone for all runs.

rxu17:
> I would vote for converting to UTC before using the actual conversion to UNIX time to have a consistency timezone for all runs.

Yup, I confirmed this is a visible issue/discrepancy in the testing, so we definitely want to. Just updated the description of this PR. This PR already handles the converting to UTC fix.

danlu1 (approved):
LGTM

PR #610, opened 2025-09-04
[GEN-1821] Provide genie user-agent string to synapse python client to track usage

danlu1:
# **Problem:**

We would like to track sites/other user’s aacrgenie usage but right now there is no user_agent information being specified in the package. 

# **Solution:**

Utilize the new functionalities from syanapseclient 4.8.0 to add `user_agent` when initiating Synapse object  

# **Testing:**

Tested that user_agent information is added successfully when synapseclient is 4.9.0.
Screenshot 2025-09-04 at 11 53 10 AM

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -30,7 +30,7 @@ project_urls =
 packages = find:
 install_requires =
     pyranges==0.1.4
-    synapseclient[pandas]>=4.0.0, <5.0.0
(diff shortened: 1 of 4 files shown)


setup.cfg
rxu17, line 1: I would add this to `requirements.txt` as well

rxu17 (approved):
LGTM! Once everything passes including integration tests

PR #611, opened 2025-09-17
[GEN-1892] redeact pediatric YEAR_CONTACT and YEAR_DEATH

danlu1:
# **Problem:**

- We initially only redacted BIRTH_YEAR, but we want to continue standardizing any center-uploaded masked data by labeling it as "withheld."

# **Solution:**

- add a function to redact year column has <. 

# **Testing:**
Unit test cases have been added.

### DIFF
--- a/genie/database_to_staging.py
+++ b/genie/database_to_staging.py
@@ -149,6 +149,22 @@ def _redact_year(df_col):
     return df_col
 
 
+def _redact_ped_year(df_col):
+    """Redacts year values that have <
+
+    Args:
+        df_col: Dataframe column/pandas.Series of a year column
+
+    Returns:
+        pandas.Series: Redacted series
+
+    """
+    year = df_col.astype(str)
+    contain_lessthan = year.str.contains("<", na=False)
+    df_col[contain_lessthan] = "withheld"
+    return df_col
+
+
 # TODO: Add to transform.py
 def _to_redact_difference(df_col_year1, df_col_year2):
     """Determine if difference between year2 and year1 is > 89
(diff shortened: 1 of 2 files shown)


genie/database_to_staging.py
```diff
@@ -149,6 +149,22 @@ def _redact_year(df_col):
     return df_col
 
 
+def _redact_ped_year(df_col):
```
rxu17: nit: add typing and return type

genie/database_to_staging.py
```diff
@@ -205,6 +221,10 @@ def redact_phi(
     )
     clinicaldf.loc[to_redact, "BIRTH_YEAR"] = "cannotReleaseHIPAA"
 
+    # redact range year for pediatric data
```
rxu17, line 227: I might have missed this in the discussion but to confirm, `YEAR_CONTACT` and `YEAR_DEATH` are also phi so that's why they are being redacted as well?

rxu17 (approved):
LGTM! Just a comment

genie/database_to_staging.py
danlu1: I think Xindi referred to the AC of the ticket:'Keep standardizing the center uploaded masked data as "withheld"' and confirmed that we need to redact these two columns. I believe these are phi but the reason to redact should be more related to standardization.

rxu17: Ah I see. Sounds good as long as that's clearly documented, see [JIRA comment](https://sagebionetworks.jira.com/browse/GEN-1892?focusedCommentId=266907)

PR #613, opened 2025-09-30
[GEN-2283] Add to validation function docstring regarding text consistency

rxu17:
# **Problem:**
For the validation function: [_check_int_year_consistency](https://github.com/Sage-Bionetworks/Genie/blob/b589b316de60e1cafaeb130ba7de020b2498b3b5/genie_registry/clinical.py#L112), it's not very clear what determines if an input clinical file has valid text consistency rows or not. 

# **Solution:**
Add to the docstring with the two scenarios that determines text consistency and provide detailed examples

# **Testing:**
On my local `mkdocs serve` it shows: 

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,14 @@ One of the features of the `aacrgenie` package is that is provides a local valid
 
 These instructions will install all the necessary components for you to run the validator locally on all of your files, including the Synapse client.
 
-1. Create a virtual environment using package manager of your choice (e.g: `conda`, `pipenv`)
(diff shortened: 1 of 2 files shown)


rxu17:
@Chelsea-Na When you get a chance, could you review this and see if it's clear enough to you?

Chelsea-Na:
This makes better sense! Thank you for providing the examples. This will make the validation rule easier to understand.

Chelsea-Na (approved):
LGTM!

PR #614, opened 2025-09-30
[GEN-1789]fix dtype not understood issue in maf validation step

danlu1:
# **Problem:**

We ran into when calling `not df.query("REFERENCE_ALLELE == TUMOR_SEQ_ALLELE2").empty`
```    
File "/usr/local/lib/python3.10/dist-packages/pandas/core/dtypes/common.py", line 1684, in pandas_dtype
      raise TypeError(f"dtype '{dtype}' not understood")
  TypeError: dtype 'COMMENTS    float64
  COMMENTS    float64
  COMMENTS    float64
  dtype: object' not understood
```
when testing 17.0 package for staging project. 

# **Solution:**

Use `any` instead to check if there are any rows meet the condition. 

# **Test:**

Unit test cases have been added to simulate duplicate columns in MSK data and all tests passed.

### DIFF
--- a/genie_registry/maf.py
+++ b/genie_registry/maf.py
@@ -113,7 +113,7 @@ def _check_allele_col_validity(df: pd.DataFrame) -> str:
     if (
         tsa2_col_exist
         and ref_col_exist
-        and not df.query("REFERENCE_ALLELE == TUMOR_SEQ_ALLELE2").empty
+        and any(df["REFERENCE_ALLELE"] == df["TUMOR_SEQ_ALLELE2"])
     ):
         error = (
             f"{error}maf: Contains instances where values in REFERENCE_ALLELE match values in TUMOR_SEQ_ALLELE2. "
--- a/genie_registry/maf.py
+++ b/genie_registry/maf.py
@@ -119,7 +119,7 @@ def _check_allele_col_validity(df: pd.DataFrame) -> str:
             f"{error}maf: Contains instances where values in REFERENCE_ALLELE match values in TUMOR_SEQ_ALLELE2. "
             "This is invalid. Please correct.\n"
         )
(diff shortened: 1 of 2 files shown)


genie_registry/maf.py
```diff
@@ -113,7 +113,7 @@ def _check_allele_col_validity(df: pd.DataFrame) -> str:
     if (
         tsa2_col_exist
         and ref_col_exist
-        and not df.query("REFERENCE_ALLELE == TUMOR_SEQ_ALLELE2").empty
+        and any(df["REFERENCE_ALLELE"] == df["TUMOR_SEQ_ALLELE2"])
```
rxu17, line 116: Do you know why using `query` and `.empty` causes the weird dtype mishap specifically with MSK data?

danlu1: As the chatGPT, it seems output multiple dtype instead of a single dtype and the evaluation errors out.

genie_registry/maf.py
rxu17, line 1: Could we add a test example for `_check_allele_col_validity` that matches what is going on with the MSK data?

danlu1: I checked and couldn't locate a input data issue for MSK data. I think it's a pandas.query issue and we can keep use the original test cases.

rxu17: Oh are you saying other input data triggers the same thing. Because the MSK dataset triggered this query issue, I'm wondering if we can add a test case that matches the MSK data that triggered this and make sure the test passes with the fix?

danlu1: I briefly checked that the MSK dataset has no difference as other cohorts. I thought it's a pandas issue instead of data issue. I will dive into it. The fix resolved the error when I was debugging.

rxu17 (approved):
LGTM!

PR #616, opened 2025-10-23
[GEN-2423, GEN-2230] DO NOT MERGE. refactor update table function, and convert pipeline to use pandas nullable types

danlu1:
**We are not merging this PR until upgrading synapseclient for GENIE package to 4.10 or newer versions.** 
# **Problem:**

The original `load.store_database` which is used in table update utilize:
1. A temporary file to append, update, and delete rows. 
2. Manually convert float value to integer.
These can be addressed easily in the `store_rows` and `delete_rows` as we upgraded synapseclient version. 

The docstring of `_create_update_rowsdf` is not clear enought. 

# **Solution:**
1. Refactored `load.store_database` to update/append rows using `store_rows`  and delete rows using `delete_rows`. 
2. Updated docstring of `_create_update_rowsdf`. 

# **Testing**
1. Unit test suites have been added. 
2. Integration test passed. See https://sagebionetworks.jira.com/browse/GEN-2230 for details.


genie/process_functions.py
```diff
@@ -557,8 +557,8 @@ def _create_update_rowsdf(updating_databasedf, updatesetdf, rowids, differentrow
     Create the update dataset dataframe
 
     Args:
-        updating_databasedf: Update database dataframe
-        updatesetdf:  Update dataset dataframe
+        updating_databasedf: Existing database dataframe with matched indices to database with updated values
```
rxu17: nit: A bit confused by the wording? What are the two objects we are matching?

genie/load.py
```diff
@@ -309,42 +307,19 @@ def store_database(
     Args:
         syn (synapseclient.Synapse): Synapse object
         database_synid (str): Synapse Id of the Synapse table
```
rxu17: nit: should we update `database_synid` to something like `table_synid` so it's more relevant?

tests/test_load.py
```diff
@@ -198,74 +197,68 @@ def test_check_database_changes(to_delete, expected_to_delete_rows):
...
+    mock_database_ent.name = "test_table"
+    # get the table entity
+    with patch.object(syn, "get", return_value=mock_database_ent) as patch_get, patch.object(Table, "store_rows") as patch_store_rows, patch.object(Table, "delete_rows") as patch_delete_rows, patch.object(load.logger, "info") as patch_info:
+        load.store_database(syn, database_synid, all_updates, to_delete_rows)
+        patch_get.assert_called_once_with(database_synid)
+        patch_store_rows.assert_not_called()
+        patch_delete_rows.assert_called_once_with(to_delete_rows)
+        patch_info.assert_called_once_with(f"Deleting {len(to_delete_rows)} rows from {mock_database_ent.name} table")
+
+def test_store_database_empty_update_and_delete(syn):
```
rxu17: nit: for easier test readability, does it make more sense to use something like `test_store_database_no_rows_to_update_and_delete` or `test_store_database_empty_update_and_delete_dataframe`. Similar comment to the rest of the tests.

genie/load.py
```diff
@@ -187,7 +188,6 @@ def _update_table(
     store_database(
         syn,
         database_synid,
-        changes["col_order"],
```
rxu17, line 190: nit: should we remove this from this as well since it's not being used: https://github.com/Sage-Bionetworks/Genie/blob/07ed7c5b528c7b72acf5542de3e5e81a8447acaa/genie/load.py#L283

genie/load.py
```diff
@@ -309,42 +307,19 @@ def store_database(
...
-    if storedatabase:
-        syn.store(synapseclient.Table(database_synid, update_all_file.name))
-    # Delete the update file
-    os.unlink(update_all_file.name)
-
+    # get the table entity
+    table_key = syn.get(database_synid)
+    # upsert table with new and updated rows
+    if not all_updates.empty:
+        Table(id = database_synid).store_rows(all_updates, to_csv_kwargs= {"float_format": "%.12g"}, schema_storage_strategy=SchemaStorageStrategy.INFER_FROM_DATA)
```
rxu17: I'm reading the API docs and wondering why the `SchemaStorageStrategy.INFER_FROM_DATA` is being used? It looks like [the case scenarios it's being used is when the table you are upserting into doesn't have columns and you're letting the data define them. ](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/e170e6f35911c899611930f4163cf89c7fc33f57/synapseclient/models/table.py#L685-L706)

rxu17 (approved):
LGTM! Thanks for this feature, I just had some comments

genie/process_functions.py
danlu1: This is to indicate two datasets are subsets of their original ones and they share matching indices (check_by in this case): https://github.com/Sage-Bionetworks/Genie/blob/a97faa26a773ee9970542f463eec6d03ada7efa1/genie/process_functions.py#L605C5-L607C1

genie/load.py
danlu1: I added here in case we have new columns in the future.

genie/process_functions.py
rxu17: I see now, the variable names being used are also super confusing. Changing all of that would be out of the scope of this. Maybe you could clarify the description to something like the following (if I'm understanding it right):

```
Args:
    updating_databasedf (pd.DataFrame):
            Subset of the existing database table data containing records
            whose checkby values also exist in the new dataset. Represents
            entries that are eligible for update or replacement based on the
            incoming dataset.
    updatesetdf (pd.DataFrame):
            Subset of the new dataset containing records whose
            `checkby` values already exist in the current database table. Represents
            entries that should be added or updated in the database table.
```
Does that represent what you are seeing?

genie/load.py
rxu17: Oh I see, new columns that are not in the current table schema so it would just auto create them?

danlu1: Yes, "The new columns created as a result of this strategy will be appended to the end of the existing columns if the table already exists." https://python-docs.synapse.org/en/stable/reference/experimental/async/table/#synapseclient.models.SchemaStorageStrategy.INFER_FROM_DATA

genie/process_functions.py
danlu1: For updatesetdf, it’s essentially updating existing rows only—no new rows are added—because _get_left_union_df isn’t a true left join/union. It matches on check_by and returns a subset of rows.
I will update the docstring.

genie/load.py
rxu17: Yup, just wanted to confirm. My concern with this is that we don't have comprehensive column validation in our input data so let's say someone names a column something and someone also has a new column called SOMETHING, those would be seen as separate things and could be let through. 

We also don't have a column validation for our processing layer so the set table schemas we have for the synapse tables are kind of that layer of validation right now if we accidentally create something in our processing pipeline.

Also for the type inference as well, it seems like it might silent coerce the type if a new dataset doesn't match the expected type in the database table. 

Unless we add validation for our processing layer, when the table schema changes, we'd only catch this at the end of a release if we do at all (even now we don't do a comprehensive comparison report at the end between releases just staging so we'd have to make that part of the QC).

This is why my preference would be for no `SchemaStorageStrategy` because of these risks and our lack of comprehensive data validation and there are other reasons as well, but just giving some major ones.

genie/load.py
danlu1: I think database_synid was used to match the function name store_database. Either works for me.

rxu17: Another option is `database_table_synid` if we want to keep it with the function name but I think the function name `store_database` also eventually needs changing because databases and tables are not the same thing.

Yeah we don't need to go change everything in this one PR but we can slowly start to migrate the naming / nomenclature to be more clear

genie/load.py
danlu1: Gotcha. I didn't know there is a layer of column validation. Thanks for shedding light. I will remove  SchemaStorageStrategy.

rxu17: Yup it will throw an error like this:
```
The first line is expected to be a header but the values do not match the names of of the columns of the table (Extra_Columns is not a valid column name or id). Header row: ROW_ID,ROW_VERSION,Chromosome,Start_Position,End_Position,Reference_Allele,Tumor_Seq_Allele2,Tumor_Sample_Barcode,Center,t_depth,t_alt_count,HGVSp_Short,Variant_Classification,Protein_position,Hugo_Symbol,inBED,FILTER,gnomAD_AF,gnomAD_AFR_AF,gnomAD_AMR_AF,gnomAD_ASJ_AF,gnomAD_EAS_AF,gnomAD_FIN_AF,gnomAD_NFE_AF,gnomAD_OTH_AF,gnomAD_SAS_AF,Annotation_Status,Extra_Columns
```

rxu17:
@danlu1 FYI, if you merge in changes from `develop`, you can use the new `run_integration_tests` label functionality: https://sagebionetworks.jira.com/wiki/spaces/APGD/pages/3562012688/Workflow+for+Adding+New+Features#Running-Test-Pipeline

danlu1:
@rxu17 Thanks, will definitely try.

rxu17:
Note to add PR to update the line that pulls from this feature branch's docker image from `scripts/sync_tables/Dockerfile` in https://github.com/Sage-Bionetworks-Workflows/nf-genie once this gets merged.

requirements.txt
```diff
@@ -5,5 +5,5 @@ httplib2>=0.11.3
 pyranges==0.1.4
 # known working version 6.0
 PyYAML>=5.1
-synapseclient[pandas]>=4.8.0, <5.0.0
```
rxu17, line 8: Could we keep the upper bound of `< 5.0.0`?

genie/load.py
```diff
@@ -308,42 +307,28 @@ def store_database(
...
-                .replace(".0\n", "\n")
+    # get the table entity
+    table_entity = syn.get(database_table_synid)
+    # upsert table with new and updated rows
+    if not all_updates.empty:
+        # TEST: convert to csv to enforce float_format="%.12g" in store_rows
+        with tempfile.TemporaryDirectory() as tmpdir:
+            tmpfile = os.path.join(tmpdir, "all_updates.csv")
+            all_updates.to_csv(tmpfile)
+            Table(id=database_table_synid).store_rows(tmpfile)
```
rxu17: Seeing an error with storing rows in this failed integration tests: https://github.com/Sage-Bionetworks/Genie/actions/runs/18978208245/job/54204110783

danlu1: This is known and I added code as Bryan suggested to hack the problem. But it's unfortunately not working. Let me test it again on my local. 
Update: float_format="%.12g" we used to do float data conversion in python client fails.

genie/process_functions.py
```diff
@@ -517,6 +517,9 @@ def _append_rows(new_datasetdf, databasedf, checkby):
         logger.info("No new rows")
     del appenddf[checkby]
     appenddf.reset_index(drop=True, inplace=True)
+    # update the ROW_ID and ROW_VERSION columns to be NaN since they are not in the original database
+    appenddf["ROW_ID"] = np.nan
```
rxu17, line 518: What issue occurs when these aren't set to `np.nan` ? This is only happening for certain table scenarios?

genie_registry/bed.py
```diff
@@ -549,8 +550,8 @@ def _process(self, beddf, seq_assay_id, newpath, parentid, create_panel=True):
         beddf["Start_Position"] = beddf["Start_Position"].apply(int)
         beddf["End_Position"] = beddf["End_Position"].apply(int)
 
-        gene_position_table = self.syn.tableQuery("SELECT * FROM syn11806563")
-        gene_positiondf = gene_position_table.asDataFrame()
+        gene_position_table = query("SELECT * FROM syn11806563")
+        gene_positiondf = gene_position_table.convert_dtypes()
```
rxu17: Does `convert_dtypes()` now automatically do the work of `asDataFrame()` and the dtype conversion?

rxu17: Nevermind I see that `query` auto converts it to a dataframe now?

genie/load.py
```diff
@@ -147,11 +148,10 @@ def update_table(
...
     """
     databaseEnt = syn.get(databaseSynId)
-    database = syn.tableQuery(
+    database = query(
         f"SELECT * FROM {databaseSynId} where {filterByColumn} ='{filterBy}'"
-    )
-    database = database.asDataFrame()
-    db_cols = set(database.columns)
+    ).convert_dtypes()
+    db_cols = set(database.columns).drop(["ROW_ID", "ROW_VERSION"])
```
rxu17: Do you know why we have to add this in now?

genie/process_functions.py
danlu1: Then the row_id and row_version will be these from the newdatasetdf which is wrong. The rows to append should be completely new without row id and version.

rxu17: Gotcha! So is this something we have to update in our database schema so we don't have to do this ? Or is this the typical workflow for any new data to upsert to a pre-existing database

genie/load.py
```diff
@@ -239,10 +241,10 @@ def _generate_primary_key(
     Returns:
         pd.DataFrame: The dataframe with primary_key column added
     """
-    # replace NAs with emtpy string
-    dataset = dataset.fillna("")
     # generate primary key column for original database
     dataset[primary_key_cols] = dataset[primary_key_cols].applymap(str)
+    # replace NAs with emtpy string
+    dataset[primary_key_cols] = dataset[primary_key_cols].fillna("")
```
rxu17: So if `applymap` converts int to float and then applies `str`, what is the goal of this later line if we don't want to add back the `.0`?

genie/process_functions.py
danlu1: It is the typical workflow for adding new rows cuz updating and appending rows are based on the underlying row_id and row_version. If your rows do not have row id and version, the client will treat it as a new row. See [here](https://python-docs.synapse.org/en/stable/reference/experimental/async/table/?h=store#synapseclient.models.Table.store_rows_async:~:text=Behind%20the%20scenese%20the%20tables%20also%20has%20ROW_ID,be%20treated%20as%20new%20rows%20to%20be%20inserted.) for details.

tests/test_load.py
rxu17, line 1: Note that for this PR and for validation of https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1272/files, just going to get these tests for the  functions that affect the [sync tables PR](https://github.com/Sage-Bionetworks-Workflows/nf-genie/pull/42/files#diff-95828994b35147247dd8158e839816eca759cb1c7bd2604e5c089b778dcdce39) to pass.

genie/load.py
```diff
@@ -222,7 +221,10 @@ def _reorder_new_dataset(
         The re-ordered new dataset
     """
     # Columns must be in the same order as the original data
-    new_dataset = new_dataset[orig_database_cols]
+    if "ROW_ID" in orig_database_cols and "ROW_VERSION" in orig_database_cols:
+        new_dataset = new_dataset[orig_database_cols.drop(["ROW_ID", "ROW_VERSION"])]
```
rxu17: I wonder why we are dropping these when we'd likely need them later when deleting rows?

genie/load.py
```diff
@@ -222,7 +221,10 @@ def _reorder_new_dataset(
         The re-ordered new dataset
     """
     # Columns must be in the same order as the original data
-    new_dataset = new_dataset[orig_database_cols]
+    if "ROW_ID" in orig_database_cols and "ROW_VERSION" in orig_database_cols:
```
rxu17: Would there ever be a scenario where the original database table did not have "ROW_ID" and "ROW_VERSION"? I'm wondering if that should even be possible / allowed. I don't think the rest of the update code would work as expected since this function is just called there. Will add logic that original database table must have "ROW_ID" and "ROW_VERSION".

We also don't want to drop ROW_ID and ROW_VERSION from the new dataset if they happen to have those columns

genie/load.py
```diff
@@ -147,11 +148,10 @@ def update_table(
         toDelete (bool, optional): Delete rows given the primary key. Defaults to False.
     """
     databaseEnt = syn.get(databaseSynId)
-    database = syn.tableQuery(
+    database = query(
         f"SELECT * FROM {databaseSynId} where {filterByColumn} ='{filterBy}'"
     )
```
rxu17, line 153: I wonder why we don't need to add `convert_dtypes()` here? When I synced the tables for example `SEQ_YEAR` is expected to be an integer dtype and have NULLs. So this uploaded to clinical database fine (since the database table expects integers only), and also when comparing, had the expected dtypes of nullable integer type.

thomasyu888: Synapse Tables accepts null values within integer columns. 

SQL nulls work different then pandas nulls that change the type of the column.

rxu17:
Closing this as won't do

PR #617, opened 2025-10-23
[GEN-2311] Deprecate dockerhub info

rxu17:
# **Problem:**
This is the start of clean-up to do for Genie repos that have migrated to using GHCR from dockerhub for docker deployment. We have already migrated to using GHCR but we still have dockerhub automatic deployment set up for this repository.

# **Solution:**
Deprecate and remove all traces of dockerhub in the documentation and switch to GHCR.

# **Testing:**
For just the `dev-container` portion verified that using VSCode's open in container feature will pull down the GHCR docker image
image

Successfully opened VSCode in the docker image, and then was able to successfully runs tests inside the dev container
image

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,5 +1,5 @@
 {
-    "image": "sagebionetworks/genie:latest",
+    "image": "ghcr.io/sage-bionetworks/genie:main",
     "mounts": [
         "source=${localEnv:HOME}/.synapseConfig,target=/root/.synapseConfig,type=bind,consistency=cached"
     ]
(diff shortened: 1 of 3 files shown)


.devcontainer/devcontainer.json
```diff
@@ -1,5 +1,5 @@
 {
-    "image": "sagebionetworks/genie:latest",
+    "image": "ghcr.io/sage-bionetworks/genie:develop",
```
rxu17, line 2: Don't think I've activated this before in VSCode development but modifying it here just in case

EDIT: Verified it works with VSCode with their open in container feature and pulls the GHCR image - see testing section above

thomasyu888 (approved):
🔥 LGTM!

PR #618, opened 2025-10-24
[GEN-2332] Add workflow to build genome nexus annotator

rxu17:
# **Problem:**
We build the Genome Nexus annotator.jar file used in our main genie maf processing pipeline manually following instructions here: https://sagebionetworks.jira.com/wiki/spaces/APGD/pages/3016687662/Updating+Genome+Nexus+Annotator+and+Dependencies#Updating-the-annotator.jar

This creates room for human error as well as no guarantee in the reproducible environment that we are using each time we manually rebuild
 
# **Solution:**
Create a github actions workflow for building the annotator.jar now that we have all of the main factors down. This will ensure consistent environment each time the annotator.jar is built. 

This workflow is meant to be triggered manually whenever we have a new commit hash for the [genome-nexus-pipeline](https://github.com/genome-nexus/genome-nexus-annotation-pipeline) that we want to use to rebuild our [annotator.jar](https://www.synapse.org/Synapse:syn22084320), we'd go to this workflow and user input the commit hash.

The goal eventually is to also add in user defined inputs for the maven and java versions as well as those are the most subject to change 2nd to the commit hash but due to the known github bug, this will have to wait.

# **Testing:**
- [x] Integration tests pass
- [x] Result of annotated test data matches previous test data (No differences found!)
- [x] Run on actual prod data example and make sure result matches previous annotated data for that center (test on WAKE)

### DIFF
--- a/.github/workflows/build_genome_nexus_annotator.yml
+++ b/.github/workflows/build_genome_nexus_annotator.yml
@@ -14,7 +14,7 @@ on:
 env:
     # versions for dependencies used in annotator.jar build
     MAVEN_VERSION: '3.9.9'
-    COMMIT_HASH: 2e67ebd08cf7c26bf1f55f2baf4b73ac36531119
+    COMMIT_HASH: ${{ github.event.inputs.commit_hash || 'default_value' }}
     JAVA_VERSION: "21"
     JAVA_DISTRIBUTION: "corretto"
     # synapse id of the Genome Nexus folder
--- a/.github/workflows/build_genome_nexus_annotator.yml
+++ b/.github/workflows/build_genome_nexus_annotator.yml
@@ -4,13 +4,6 @@ on:
   push:
     paths:
       - '.github/workflows/build_genome_nexus_annotator.yml'
-  workflow_dispatch:
-    inputs:
-      commit_hash:
-        description: "Genome Nexus pipeline commit hash to use for version of annotator.jar"
-        required: true
-        default: "2e67ebd08cf7c26bf1f55f2baf4b73ac36531119"
-
 env:
     # versions for dependencies used in annotator.jar build
     MAVEN_VERSION: '3.9.9'
--- a/.github/workflows/build_genome_nexus_annotator.yml
+++ b/.github/workflows/build_genome_nexus_annotator.yml
@@ -1,0 +1,119 @@
+name: Build and Upload Genome Nexus Annotator
+
+on:
+  workflow_dispatch:
+    inputs:
+      commit_hash:
+        description: "Genome Nexus pipeline commit hash"
+        required: true
+        default: "main"
+
+jobs:
+  build-annotator:
+    runs-on: ubuntu-latest
+
+    env:
+      MAVEN_VERSION: 3.9.9
+      SYNAPSE_ENTITY_ID: syn22084320
+      SYNAPSE_AUTH_TOKEN: ${{ secrets.SYNAPSE_AUTH_TOKEN }}
+      PROD_DOCKER: ghcr.io/sage-bionetworks/genie:main
+      TEST_PROJECT_SYNID: syn7208886
+
+    steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/build_genome_nexus_annotator.yml
```diff
@@ -0,0 +1,119 @@
...
+
+      - name: Upload annotator.jar to Synapse
+        env:
+          COMMIT_HASH: ${{ github.event.inputs.commit_hash }}
+        run: |
+          cd genome-nexus-annotation-pipeline
+          FILE_PATH=$(find . -name annotator.jar | head -n 1)
+          echo "Uploading $FILE_PATH to Synapse entity $SYNAPSE_ENTITY_ID..."
+
+          synapse login --authToken "$SYNAPSE_AUTH_TOKEN"
```
thomasyu888: The auth token param may not be needed due to the environment variable

thomasyu888 (approved):
🔥 this is really nice! Great work here.

.github/workflows/build_genome_nexus_annotator.yml
```diff
@@ -0,0 +1,139 @@
+name: Build and Upload Genome Nexus Annotator
+
+on:
+  push:
+  workflow_dispatch:
```
rxu17: There is a known bug with Github where the workflow dispatch doesn't show up unless you rename it (even then it's not guaranteed?): https://github.com/orgs/community/discussions/25219 even when your workflow is checked into the default branch of your repo. We can tackle once we get there but for now I'll make it so you can trigger this by a change to the github yaml file.

.github/workflows/build_genome_nexus_annotator.yml
```diff
@@ -0,0 +1,141 @@
...
+    inputs:
+      commit_hash:
+        description: "Genome Nexus pipeline commit hash to use for version of annotator.jar"
+        required: true
+        default: "2e67ebd08cf7c26bf1f55f2baf4b73ac36531119"
+
+env:
+    # versions for dependencies used in annotator.jar build
+    MAVEN_VERSION: '3.9.9'
+    COMMIT_HASH: 2e67ebd08cf7c26bf1f55f2baf4b73ac36531119
```
danlu1: Can this be set to something like: COMMIT_HASH: ${{ github.event.inputs.commit_hash || 'default_value' }} so there is only one place to hold the commit_hash?

rxu17: Thanks! I think I forgot the `'default_value'` last time and that's why it didn't work

rxu17: Actually I remembered why it doesn't work. Right now it will only have that `default` that i setup if I ran it through the github actions UI but because of the bug I mentioned, we can't test that until later so I just have the 
```
workflow_dispatch:
Collapse commentComment on line R7rxu17 commented on Oct 24, 2025 rxu17on Oct 24, 2025ContributorAuthor
More actionsThere is a known bug with Github where the workflow dispatch doesn't show up unless you rename it (even then it's not guaranteed?): https://github.com/orgs/community/discussions/25219 even when your workflow is checked into the default branch of your repo. We can tackle once we get there but for now I'll make it so you can trigger this by a change to the github yaml file.
React
    inputs:
      commit_hash:
        description: "Genome Nexus pipeline commit hash to use for version of annotator.jar"
        required: true
        default: "2e67ebd08cf7c26bf1f55f2baf4b73ac36531119"
```
set up for later when we are able to implement the fix when it gets pushed to `main` but right now, just push a commit to trigger this workflow. Although now that I'm thinking about it, we probably want to update the commit hash via PR to build the annotator.jar instead of doing it via `workflow_dispatch:` because that doesn't save our default commit value that we pick if it changes (and eventually it will change)

danlu1 (approved):

PR #619, opened 2025-10-31
[GEN-1654] Add fix to CI/CD for build docker image

rxu17:
# **Problem:**
Our workflow will only trigger the CI/CD to build docker images and run integration tests if it's within a PR and has label `integration_tests` for feature branches (otherwise it triggers on push events for develop and main branches)

But there's an if statement that checks for a non-PR event to build and push the docker image which makes it impossible to build docker images for feature branches since we disabled that for the CI/CD for feature branches

We also need to tag the branch name instead of using the PR name when building the docker image.

# **Solution:**
- Remove impossible to access if statement from build docker image
- Use `github.head_ref` when in PR to build docker image from feature branch instead of PR name, use `github.ref_name` otherwise

# **Testing:**
Build here: https://github.com/Sage-Bionetworks/Genie/actions/runs/18965295444 should run successfully and build the entire docker image

Build is happening:
image

Successful building and pulling of correct image name: 
image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -184,7 +184,6 @@ jobs:
       - name: Build and push Docker image
         id: build-and-push
         uses: docker/build-push-action@v5
-        if: github.event_name != 'pull_request'
         with:
           context: .
           push: true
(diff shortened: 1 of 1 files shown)


danlu1 (approved):
LGTM!!

rxu17:
Blocking #616

PR #620, opened 2025-10-31
[GEN-2375] Replace deprecated `DataFrame.append` with `pandas.concat`

rxu17:
# **Problem:**
We ran into an error with 19.4 consortium release at the dashboard update step where it calls `DataFrame.append` but since we upgraded to pandas 2.3.1, that method is now deprecated since ~pandas 2.0.

[See release notes for 2.0 ](https://pandas.pydata.org/pandas-docs/version/2.0/whatsnew/v2.0.0.html)

> Removed deprecated Series.append(), DataFrame.append(), use concat() instead

JIRA Ticket: https://sagebionetworks.jira.com/browse/GEN-2375

# **Solution:**
Swap to using `pd.concat`. Note that this will be a PR to `main` so we can finish the consortium release.

# **Testing:**
- Unit tests pass
- There is unfortunately no staging or testing mode for the dashboard updater because we don't have many of those tables yet

### DIFF
--- a/genie/dashboard_table_updater.py
+++ b/genie/dashboard_table_updater.py
@@ -51,18 +51,26 @@ def get_center_data_completion(center, df):
 
 
 def update_samples_in_release_table(
-    syn, file_mapping, release, samples_in_release_synid
-):
+    syn: synapseclient.Synapse,
+    file_mapping: dict,
+    release: str,
+    samples_in_release_synid: str,
+) -> None:
     """
     Updates the sample in release table
     This tracks the samples of each release.  1 means it exists, and 0
-    means it doesn't
+    means it doesn't. For releases without a column in the sample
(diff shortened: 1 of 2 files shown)


genie/dashboard_table_updater.py
```diff
@@ -459,9 +467,9 @@ def update_sample_difference_table(syn, database_mappingdf):
         .applymap(int)
     )
 
-    diff_between_releasesdf[
-        ["Clinical", "Mutation", "CNV", "SEG", "Fusions"]
-    ] = new_values
+    diff_between_releasesdf[["Clinical", "Mutation", "CNV", "SEG", "Fusions"]] = (
```
rxu17: Just lint change

thomasyu888 (approved):
🔥 LGTM!

PR #621, opened 2025-11-01
[GEN-2377] Remove set as index

rxu17:
# **Problem:**
We are running into an issue in the dashboard updater step where you can no longer use `set`s in pandas as the index of a dataframe. 

JIRA Ticket: https://sagebionetworks.jira.com/browse/GEN-2377

# **Solution:**
Similar fix to https://github.com/Sage-Bionetworks/Genie/pull/554, where we remove `set` as the index determinant, but instead add `sorted(set())` as that's more robust than `list(set())` and gives you a guaranteed order each time rather than leaving it up to `set`

# **Testing:**
- Unit tests pass
- There is no testing or staging mode set for this so it can't be tested in integration testing at the moment


thomasyu888 (approved):
🔥 LGTM!

PR #622, opened 2025-11-04
[GEN-2324] Add VCF validation for expected columns

rxu17:
# **Problem:**
We need to add VCF validation for:
- expected `FORMAT` column
- expected tumor/sample columns given the number of columns present in the VCF itself
- expecting NO NAs in the `FORMAT` or tumor/sample columns

JIRA Ticket: https://sagebionetworks.jira.com/browse/GEN-2324
# **Solution:**
New `validate_tumor_and_normal_sample_columns_exist` function to do all of the above. Because the validation of the tumor/sample columns are very coupled together with validating their actual values, it was easier to keep the validation together in one function.

# **Testing:**
- [x] - Unit tests pass
- [x] - Integration tests pass
- [x] - Add new input vcf data that would fail and produce certain errors

Data modified:

Added missing value to `GENIE-SAGE-1-1` column, got expected validation error message
```
You have invalid files! Here are the reasons why:

Filenames: GENIE-SAGE-1-1.vcf, Errors:
 ----------------ERRORS----------------
vcf: Must not have missing values in GENIE-SAGE-1-1 column.
```
Made the `FORMAT` column called `FORMAT2` in input data, got expected validation error message
```
You have invalid files! Here are the reasons why:

Filenames: GENIE-TEST-1-1.vcf, Errors:
 ----------------ERRORS----------------
vcf: Must have these headers: CHROM, POS, ID, REF, ALT, QUAL, FILTER, INFO, FORMAT.
```

### DIFF
--- a/genie_registry/vcf.py
+++ b/genie_registry/vcf.py
@@ -133,9 +133,7 @@ class vcf(FileTypeFormat):
                 )
             if vcfdf["FORMAT"].isnull().values.any():
                 total_error += "vcf: Must not have missing values in FORMAT column.\n"
-        total_error += self.validate_tumor_and_normal_sample_columns_exist(
-            input_df=vcfdf
-        )
+        total_error += self.validate_tumor_and_normal_sample_columns(input_df=vcfdf)
         # Require that they report variants mapped to
         # either GRCh37 or hg19 without
         # the chr-prefix.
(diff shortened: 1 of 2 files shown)


genie_registry/vcf.py
```diff
@@ -215,3 +174,101 @@ def _validate(self, vcfdf):
         # will parse a VCF in a detailed way,
         # and output with warnings or errors if the format is not adhered too
         return total_error, warning
+
+    def validate_tumor_and_normal_sample_columns_exist(
```
rxu17: rename to something more appropriate like `validate_tumor_and_normal_sample_columns`

genie_registry/vcf.py
```diff
@@ -215,3 +174,101 @@ def _validate(self, vcfdf):
...
+        else:
+            # Must have a column called TUMOR or sample column in the header
+            error = (
+                "vcf: Must have at least 10 columns. "
+                "If the vcf represents a single sample, then it's missing a tumor sample column. "
+                "If the vcf represents a matched tumor normal, then it's missing both normal sample and tumor sample columns.\n"
+            )
+
+        # validate the values in the tumor and/or normal sample columns if present
+        if sample_id:
```
rxu17: nit: Use `process_functions.checkColExist` instead

tests/test_vcf.py
rxu17, line 1: Add test case for when `FORMAT` has at least one NA value

tests/test_vcf.py
```diff
@@ -325,3 +333,160 @@ def test_that__get_dataframe_uses_correct_columns_to_replace(
...
+        "only_nine_cols",
+        "valid_tumor_sample",
+        "invalid_tumor_sample_name",
+        "tumor_sample_col_has_nas",
+        "valid_matched_tumor_normal",
+        "invalid_normal_sample_name",
+        "more_than_11_cols",
+    ],
+)
+def test_validate_tumor_and_normal_sample_columns_exist(
```
rxu17: We really depend on the sites to have the tumor and normal columns in a specific order (1st tumor, then normal) and there's no other way to verify if one of the columns is tumor vs normal on just their column names. 

Generally based on past exp, sites using standard vcf files will always have the above, so this is OK for now

genie_registry/vcf.py
```diff
@@ -215,3 +172,97 @@ def _validate(self, vcfdf):
...
+            least a tumor sample column on top of the 9 required VCF columns
+            - If tumor sample and/or normal sample columns are present, they must not have
+            any missing values.
+
+        Example: VCF with Matched Tumor Normal columns
+            | OTHER_VCF_COLUMNS | GENIE-GOLD-1-1-tumor | GENIE-GOLD-1-1-normal |
+            | ----------------- | -------------------- | --------------------- |
+            |        ...        |         ...          |          ...          |
+
+        Example: VCF with Single Tumor VCF column
```
danlu1: Do you mean single sample vcf here?

tests/test_vcf.py
```diff
@@ -38,6 +38,8 @@ def test_validation_valid_no_samples(vcf_class):
             "QUAL": ["AAK1", "AAED1", "AAAS"],
             "FILTER": ["AAK1", "AAED1", "AAAS"],
             "INFO": ["AAK1", "AAED1", "AAAS"],
+            "FORMAT": ["AAK1", "AAED1", "AAAS"],
+            "TUMOR": ["AAK1", "AAED1", "AAAS"],
```
danlu1: This test is duplicate with the below one sample one. As the Jira ticket, for single sample vcf, we check the existence of either Tumor or Sample column. I would update the one sample test with sample column.

danlu1 (approved):
LGTM, just some minor comments.

tests/test_vcf.py
rxu17: Actually since this is just for testing the specific tumor/normal validation scenario on the entire _validate function and we already have other tests, I'd just remove this because we already have tests for all of the individual scenarios for the specific tumor/normal validation function

genie_registry/vcf.py
rxu17: Yeah actually the nomenclature can be both ways VCF with single sample column or VCF with single tumor column (single sample), I'll make both examples

PR #629, opened 2025-12-12
[GEN-2326] Centralize where error reports get saved in Genie

rxu17:
# **Problem:**
Originally, our validation error reports for sites were being saved to their center's staging folders. This is not a very intuitive place to save it as the validation step of the pipeline is earlier in the pipeline compared to staging which is later.

# **Solution:**
Our maf error reports get saved to an `errors` folder, we should have validation errors saved to that folder as well. We should also name the report as `_validation_errors.txt` instead of `_errors.txt` so it's clear what errors these are

# **Testing:**

- [x] Unit tests pass
- [x] Integration tests pass and now test sites' errors get saved to errors folder

Validation errors are saved successfully: 
image

### DIFF
--- a/genie/write_invalid_reasons.py
+++ b/genie/write_invalid_reasons.py
@@ -30,20 +30,19 @@ def write(
     center_errors = get_center_invalid_errors(syn, error_tracker_synid)
     for center in center_mapping_df["center"]:
         logger.info(center)
+        center_errors_report_name = f"{center}_validation_errors.txt"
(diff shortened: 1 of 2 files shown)


thomasyu888 (commented):
🔥 LGTM, but i'll defer to @danlu1 for final review

thomasyu888 (approved):

danlu1 (approved):
LGTM!

PR #630, opened 2025-12-18
[GEN-2368] Deprecate meta_fusions.txt

rxu17:
# **Problem:**
Outstanding item from [old epic to depreciate fusions fileformats](https://sagebionetworks.jira.com/browse/GEN-471). Forgot to deprecate the metadata file for the fusions fileformat (deprecated).

JIRA Ticket: https://sagebionetworks.jira.com/browse/GEN-2368

# **Solution:**
Note that this will prevent links from being created in future releases that point to the deprecated `meta_fusions.txt` file. The original file will remain but mainly for historical purposes.

# **Testing:**
Our test project only has one release version that we work off of. 

In the consortium release, the link to the `meta_fusions.txt` is not updated on 12/18
image

Similar with public release:
image

### DIFF
--- a/genie/consortium_to_public.py
+++ b/genie/consortium_to_public.py
@@ -5,7 +5,6 @@ import os
 import sys
 
 import pandas as pd
-import synapseclient
 import synapseutils
 from genie import (
     create_case_lists,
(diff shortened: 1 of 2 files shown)


thomasyu888 (approved):
🔥 LGTM!

PR #631, opened 2025-12-20
[GEN-2330] Remove `includeInPanel` from our genie releases

rxu17:
# **Problem:**
We want to deprecate `includeInPanel` from our genie releases - public and consortium. We will still keep it in our internal tables but we will no long require it from sites.

Related PRs: https://github.com/Sage-Bionetworks-Workflows/nf-genie/pull/50
JIRA TIcket: https://sagebionetworks.jira.com/browse/GEN-2330

# **Solution:**
- Remove `includeInPanel` from our gene panel creation, our validation requirement and our exported fields in the bed file in our releases (genomic_information.txt). 
- Also remove the filtering we do in our dashboarding and gene panels so that we include all genes (do not filter on `includeInPanel` anymore)
- Create placeholder `includeInPanel` in our processing when it isn't present.

# **Testing:**
- Unit tests
- Integration tests pass
   - When uploading a bed file without a `includeInPanel` column, validation still passes and bed file gets processed all the way through
   - When uploading a bed file with a `includeInPanel` column, validation still passes and bed file gets processed all the way through
   -  Final consortium released bed file contains no `includeInPanel` column
   - Final public released bed file contains no `includeInPanel` column

### DIFF
--- a/genie_registry/bed.py
+++ b/genie_registry/bed.py
@@ -516,8 +516,8 @@ class bed(FileTypeFormat):
         seq_assay_id = seq_assay_id.replace("_", "-")
 
         # add in placeholder includeInPanel column and value if it doesn't exits
-        if len(beddf.columns == 4):
-            beddf[len(beddf.columns)-1] = True
+        if len(beddf.columns) == 4:
+            beddf[4] = True
         
         # Add in 6th column which is the clinicalReported
         if len(beddf.columns) > 5:
--- a/genie_registry/bed.py
+++ b/genie_registry/bed.py
@@ -456,9 +456,12 @@ class bed(FileTypeFormat):
         """
         LOGGER.info("CREATING GENE PANEL")
         if not beddf.empty:
-            exonsdf = beddf[beddf["Feature_Type"] == "exon"]
+            # exonsdf = beddf[beddf["Feature_Type"] == "exon"]
             # Only include genes that should be included in the panels
-            include_exonsdf = exonsdf[exonsdf["includeInPanel"]]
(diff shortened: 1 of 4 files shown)


thomasyu888 (approved):
🔥 LGTM! Thanks for pushing this through

xindiguo (approved):

genie_registry/bed.py
```diff
@@ -513,6 +516,10 @@ def _process(self, beddf, seq_assay_id, newpath, parentid, create_panel=True):
         seq_assay_id = seq_assay_id.upper()
         seq_assay_id = seq_assay_id.replace("_", "-")
 
+        # add in placeholder includeInPanel column and value if it doesn't exist
```
rxu17, line 519: There's a bit of a complication. If we allow sites to not submit `includeInPanel`, they could technically submit `clinicalReported` as well which is optional for now (but sites do submit data for this and it is present in our data guide) and both columns have TRUE, FALSE values. 

This creates complication for us to create a placeholder `includeinpanel` column because how do we know if a column was for `includeInPanel` or `clinicalReported` or something else? I guess technically right now, something could also happen without this in place - a site could accidentally just submit a `clinicalReported` column but our processing would auto-assume it's `includeInPanel` because our bed files are column order dependent and valid it as such.

Because bed files don't allow headers, I think we have to strictly enforce the number of columns that sites can submit OR force sites to keep includeInPanel in but just as a placeholder column

rxu17: Per internal discussion, we'd want to consider the option to enforce removing both `includeInPanel` and `clinicalReported` fields (not allowing them). This PR will be closed for now and re-opened once that discussion is complete.

PR #636, opened 2026-02-12
[GEN-2537] Storing processed maf file first then upload to maf db

rxu17:
# **Problem:**
Often we run into the maf data hitting schema limitation errors but without the data saved to troubleshoot, it makes it a lot harder especially with centers that have hundreds or thousands of mutation data.

JIRA: https://sagebionetworks.jira.com/browse/GEN-2537

# **Solution:**
We want to save the center maf file first before uploading the maf database. 

Extra: Remove call to `check_annotation_error_reports` because that is a big query call / reduce time and the failed annotations in the annotation report and what is in the maf database may not match up because sometimes there are additional failures that would not be recorded in the failed annotations report generated by genome nexus. 

# **Testing:**
- Unit tests pass
- Integration tests pass

### DIFF
--- a/genie/process_mutation.py
+++ b/genie/process_mutation.py
@@ -497,10 +497,11 @@ def split_and_store_maf(
         narrow_maf_chunk = maf_chunk[narrow_maf_cols]
         append_or_createdf(narrow_maf_chunk, annotation_paths.narrow_maf_path)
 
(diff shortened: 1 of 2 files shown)


thomasyu888 (approved):
🔥 LGTM!

PR #637, opened 2026-02-12
[GEN-2538] Fix dockerbuild with missing module `pkg_resources`

rxu17:
# **Problem:**
Docker build from scratch is running into a `ModuleNotFoundError: No module named 'pkg_resources'` during `synapser` installation when building the renv part of the docker build. This is due to version 82 of `setuptools` removing `pkg_resources`: https://github.com/pypa/setuptools/issues/5174

JIRA: https://sagebionetworks.jira.com/browse/GEN-2538

Note this bug will only affect new docker builds not `develop` or `main` builds as those are cached (if it's only code changes) so it will prevent feature branch docker builds

# **Solution:**
Force the reticulate python environment to install `setuptools` to a pinned version before restoring the R packages. `pkg_resources` is used by `synapser`. 

# **Testing:**
- Integration tests pass
- Comparison reports between develop and this feature branch run yield no differences

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -64,11 +64,14 @@ WORKDIR /root/Genie
 # a long time and unless there are changes to the actual
 # R packages used.  So the only files copied over are
 # renv/ renv.lock and the installation R script
-# Make a dedicated venv for reticulate
+# Make a dedicated venv for reticulate and force the pinning of setuptools
 ENV RETICULATE_PYTHON=/opt/reticulate-venv/bin/python
+
 RUN python3.10 -m venv /opt/reticulate-venv \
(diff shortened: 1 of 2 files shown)


rxu17:
Let's not merge this until package release is complete.

rxu17:
Just realized this will need to be included as part of this release in order to get the package specific docker image to build (because that one builds from scratch) otherwise it won't work

thomasyu888 (approved):
🔥 LGTM!   Just approving in case you needed approval, and Dan can loop back around if needed

danlu1 (approved):
LGTM

PR #642, opened 2026-06-09
Update `requests >= 2.34.0` due to security vulnerabilities

rxu17:
# **Problem:**
Currently our `requests` library is not anchored to a specific version and seems to be currently ~2.32
See security issues brought up via this PR: https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1385

# **Solution:**
Lock down `requests` to be >= 2.34 or higher

Note: This ticket also locks down `chardet` to the current working version that's in our main genie production image due to an issue raised here with newer versions of chardet: https://sagebionetworks.jira.com/browse/GEN-2592

# **Testing:**
- [x]  Test run of pipeline still works with newly built docker image with higher `requests`
- [x] Additional tests that tests functions that use `requests`

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
-# known working version 4.0.0
-chardet>=3.0.4
+# locked down chardet due to encoding issues with newer versions
+chardet>=3.0.4, <=5.2.0
 # known working version 0.20.4
 httplib2>=0.11.3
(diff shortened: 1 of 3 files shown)


thomasyu888 (approved):
🔥 LGTM! cc @danlu1 for final review

PR #643, opened 2026-06-09
[GEN-2631] Increase determine data type limit for mutation files

rxu17:
# **Problem:**
See JIRA ticket for more info but we ran into ` ValueError: Integer column has NA values in column 5` error due to the following columns having NAs past the 5000 row limit (NAs starting around ~7000 rows) used 

- t_alt_count
- t_depth
- t_ref_count

JIRA Ticket: https://sagebionetworks.jira.com/browse/GEN-2631

# **Solution:**
Increase this row limit to 10,000 in determining data type to support mutation files that have more than 5000 rows of non-NA values

# **Testing:**
- [x] Test pipeline runs
- [x] Comparsion report yields the same results before and after
- [x] Unit tests added

### DIFF
--- a/genie/process_mutation.py
+++ b/genie/process_mutation.py
@@ -16,6 +16,9 @@ from . import extract, load, process_functions
 
 logger = logging.getLogger(__name__)
 
+# Change this nrows to 10000 so that it better encapsulates the types
+MAX_ROWS_TO_DETERMINE_DTYPE = 10000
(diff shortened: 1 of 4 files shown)


Chelsea-Na (approved):
lgtm!

thomasyu888 (approved):
🔥 LGTM!

PR #644, opened 2026-06-16
[GEN-2631] Add certain columns to known float columns

rxu17:
# **Problem:**
Despite setting the number of rows snippet when sampling a maf dataset to determine data types of that column to 10k (https://github.com/Sage-Bionetworks/Genie/pull/643), we are still encountering data that have 10k rows in a row without a NA present triggering the same error.

JIRA Ticket: https://sagebionetworks.jira.com/browse/GEN-2631

# **Solution:**
Given we can't control the number of NAs that can appear in a 10k row snippet when sampling a maf dataset to determine data types of that column, let's just convert columns that should use the `float` dtype to do so so it doesn't convert to `int` (since pandas `int` can't support NAs and we currently can't support pandas nullable dtypes which do support NAs)

# **Testing:**
- Unit tests pass
- Integration tests pass
- No differences found in files


thomasyu888 (approved):
🔥 LGTM!

thomasyu888 (commented):
Are there memory implications that make us have to switch back to 5000 lines or is 10,000 fine?

rxu17:
> Are there memory implications that make us have to switch back to 5000 lines or is 10,000 fine?

I took a look at VICC (center with a lot of mutation files) and it doesn't look like memory usage changed but maybe CPU usage? (20.5 vs 20.4) Other centers doesn't seem like there's a trend:

**20.5 consortium run for VICC**
image

**20.4 consortium run for VICC**
image

rxu17:
Build is failing with `Error: failed to retrieve package 'AnnotationDbi@1.60.0'`. This happened before, and re-running seemed to have fixed it. Will wait a day or so and try again

EDIT: Build is working again

